PR #344, opened 2014-10-16
Added Feature To Automattically take game end screenshots.

triscal2000:
Also re-ordered some of the xaml to make it a bit more logical


andburn:
I've built upon this and made a plugin to the same thing. See [here](https://github.com/andburn/hdt-plugin-endgame).

PR #571, opened 2015-01-11
Added the ability to save to an own data directoy

becelot:
With every new release, I was constantly updating the Config.Instance.DataDir variable to point to my cloud storage, such that I had access to all current decks from different devices. 

I finally decided to to add the ability into the options menu, such that the folder can be set and saved in the config file. It can be found in options > Other > Tracker at the very bottom

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -67,7 +67,10 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue(-1)]
 		public int CustomWidth = -1;
 
-		[DefaultValue(false)]
+        [DefaultValue("")]
+        public string DataDirPath = string.Empty;
+
+        [DefaultValue(false)]
 		[XmlIgnore]
 		public bool Debug = false;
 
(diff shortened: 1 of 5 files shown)


Epix37:
This is currently not quite working with the way the files are copied when disabling "Save Data in AppData". It just checks and moves them from/to "locally", not the DataDir path. 

(Also I realized replays are currently not moved at all)

becelot:
yea, forgot about the replays. Will look into copying the files, haven't checked that.

Epix37:
I think that happens in MainWindow_Load: SetupDeckListFile(), SetupDeckListFile() and SetupDeckStatsFile().

becelot:
Copying should now work properly.

becelot:
still buggy, missed a case

becelot:
Fixed, should work now

Epix37:
Crashes on start when turning off "Save Data in AppData" without having a path set. 

`dataDirPath` is `"\\[file].xml"` at that point.

Also: what  happens if I have "Save Data in AppData" disabled and then change `DataDirPath`? I actually don't know but I would guess it's not handled.

becelot:
Fixed this. The second one was already implemented and handled. The first one is probably due to some changes in the WinAPI. I'm currently using Windows 10 preview. It seems, that "\\[file].xml" references the local path in Windows 10, but not in the older versions. The config has been updated, such that the standard value of DataDirPath is set to "." now, since ".\\[file].xml" should be a valid path in the older versions. Could you double check this?

Epix37:
Looks good :). I fixed the last few issues I found.

PR #598, opened 2015-01-21
Support elitedecks.net

skaforey:
Add support for importing decks from elitedecks.net


Epix37:
Looks good, thank you!

PR #599, opened 2015-01-22
Deck versioning

MarioZG:
I have added something that was bugging me for some time - deck versioning. You can now change few cards, and instead creating new deck, save new version and keep old one for reference. Stats for both versions will accumulate, but you can see also stats for each version separately. 

Most of the code and changes are done "to make it work for myself". I just thought it might be also useful for others.


Epix37:
I really like the idea but it feels.. confusing?

A few issues I see right now:
* Without having any changes, there's just an empty panel on the right, not at all telling you what it's supposed to be. Maybe there should even be an extra button to expand this?
* The version indicator in edit mode should probably not be in the menu bar.
* Yet another dialog after saving a edited deck ("I just wanna edit a card, leave me alone!").
* That -1! lol (we need to clean up the numbers in general though)
* The history panel  still contains the last edited deck when creating a new one or editing one without a history.
* Stats overview needs some changing, height increase mainly.
* Version numbers in help and the tray tooltip now display "0.7.-1"

What if we moved the version indication and increment to the top of the history panel? That would pretty much solve the first 3 of the problems. 
Something alone the lines of "[Textbox with current version] -> [Combobox containing current, minor increment and major increment]"?

GiggliG:
I cant comment on how this works, but the idea is something I find EXTREMELY useful.

Epix37:
Hehe, I figured *you* might like this one!

MarioZG:
I didn't spend too much time with UI, just made it work and displayable. If you think this might get pulled to master, i'll tidy this up.  

In terms of -1 icon, that sums up my icon design skills :).  I cannot guarantee anything much better than that.

As for additional message box during save - i had button initially to increase version, but I always forgot to click it before save, hence message box.  Drop down idea might work, with 3 options as in dialog (Major/Minor/None). None by default,  and when deck is edited switch it to major.

Epix37:
I would love to merge this into master, yes! 

If you need any help fixing things up, let me know. I should have some spare time this weekend.

MarioZG:
I should have some time later today/tomorrow to tidy UI - keep an eye for new commits ;)

MarioZG:
I have polished this a little bit.
 - now it's just major version (1,2,3....)
 - added tickbox next to save "increase version"
 - improved history panel display
 - changed layout of deck stats. Works same way as overall stats, with separate tab for game list
 - added deck version to deck's game list
 - fixed version issue in help and  try icon
 - -1 icon is still the same :), can't work the graphics... 

Added, when moving game to other deck, game is assigned to current version of target deck

Fell free to do any UI changes - i believe you have better feel for it ;)

Epix37:
Pushed this to [feature/DeckVersioning](https://github.com/Epix37/Hearthstone-Deck-Tracker/tree/feature/DeckVersioning) to at least have the merge conflicts out of the way. 

Looks a lot better now. Still needs quite a bit of work but we're getting there :).

I also liked having major and minor, I think I'm gonna put that back in.

Epix37:
How about graying out the cards instead of having red numbers?

screenshot 2015-01-24 09 33 52

Epix37:
The history panel now matches the rest a bit better, reintroduced minors and changed the version selection a bit:

screenshot 2015-01-24 10 28 26

MarioZG:
That looks much better now.

I would move dropdown with version selection next to save button. They seems to belong together, since they are part of save process.

Epix37:
You've got a point. Like this maybe? 
save

MarioZG:
Exactly. Looks perfect.

Epix37:
If anyone is up for some testing: http://www.mediafire.com/download/ms6zl7ai66c2xru/Hearthstone_Deck_Tracker-v0.7.6_PRE5.zip

(@GiggliG   :))

GiggliG:
lol, I'm up for testing.
If i install it in a separate folder, will it use its own data store, and is there a way I can import a COPY of my old data into it so it doesn't change that data in use? I'm worried about data corruption.

Epix37:
Easiest will be to just create a backup of you appdata folder. I went back and forth between this and v0.7.5 a few times without problems, should be fine. 

Just to be save, probably still a good idea.

GiggliG:
Okay, backed that up and will test this over the weekend.

GiggliG:
I've done a quick test, and it looks good so far.
I made a deck change in HDT, and saved it as v1.1.
I then played a game in hearthstone, and realised I'd forgotten to make the change the deck there. That game then got saved as a v1.1 game, though it was still using the v1.0 deck list.

So I think you'll need a button to "Move Game to other version of this deck" like you have with "Move to other deck". I think it makes sense for these to be two different buttons, and the "change version" button should be greyed out if the active deck only has one version.

Some feature suggestions:

I'd like the (optional, not required) ability to put a text descriptor with a version name, like
Secrets Mage, v1.1, Anti-aggro

I think you could have a Version Switcher, at the top of the deck display, so you can see what versions exist, and change which one is active and displayed. This would not be in edit mode, but in normal mode.

This would allow people to flip through their versions, and also change which one HDT detects as the active one when playing hearthstone. I'm surer there are people who will create multiple versions, and then go back to an earlier version as the most effective one.

If you have this view version mode, you could have a popup or panel showing the differences between the current view, and the last version that was active - like on hearthpwn, where they have updates, like "+1 Knife Juggler, -1 Cold Light Oracle" or whataver"
It could be a hovering tooltip somewhere , maybe, to avoid clutter, or could be a simple text added to the bottom of the deck list.

Here's one out-there feature suggestion that I think will be very useful for some (like me0 but maybe not for most, so i could understand if you skip this.
Have an option for the program to treat major versions (v1, v2, v3) as separate decks, which each can appear as choices in the deck detection/selection screen when you play in hearthstone.
For example, I know some people maintain multiple versions of a deck that are almost identical, but switch between them depending on changes in the meta. I maintain two versions of several decks for playing in the US and EU servers, which are very similar, but have differences because my card collections aren't the same in those two servers. I'd love to be able to combine them into a single deck, and have HDT suggest the two major versions when I start a game (or use the correct one, when it detects the cards that only exist in that deck).

GiggliG:
I'll continue testing, as I play more games :)

GiggliG:
Another suggestion: with this change, all decks get (v1.0) shown as part of their name in the deck list.
I suggest having that only show up if there is more than one version. Allowing people to see at a glance which of their decks have multiple versions.

Will you also be integrating versions with the "Update from Web" feature? So that if a deck has changed online, the original version is saved, and a new version is added with the changes?

Epix37:
"Update from web" should work just as any editing would. Or am I misunderstand you somehow?

MarioZG:
In feature branch, there is [XmlIgnore] Attrbute on PlayerDeckVersion, which  means that deck version is not stored with game. All stats are assigned to v 1.0 when you start application.

Epix37:
Oops, thanks.

MarioZG:
If you save with same version, current deck is added to history. Fix: https://github.com/MarioZG/Hearthstone-Deck-Tracker/commit/aa9557ced64205d3c114679e9d8609ca9ca801ae

GiggliG:
Regarding "Update from web": Every time I've used it, it has been "The deck is up to date" so I'm not sure what happens when it does find changes to update. So maybe my suggestion on that isn't needed.

GiggliG:
One thing I've noticed.
In the stats section, it shows games broken down by version, like v1.0, v1.1.
However, games from before the version feature was added are not listed in the first version. They are in ALL, but should be included in the first version of the deck too.

Epix37:
Okay, I think all the minor things (mentioned) are fixed. Being able to select the version is a good idea, I will try to imlement that. 

Selecting from the majors (or even all versions, for what I care :)) is a neat idea but probably won't make it into the first release of this.

Epix37:
Oookay, that was a lot harder than expected but it seems to be working somewhat. 

Latest build with the fixes and version selection: http://www.mediafire.com/download/42hyw1s8sw42b8b/Hearthstone_Deck_Tracker-v0.7.6_PRE6.zip

GiggliG:
I haven't tried the new version yet, but I have one or two new suggestions:

When you save a deck, you get the dropdown: Save 1.0, 1.1, 2.0 etc. I suggest adding to the bottom of this: "New Deck" (or having a separate button: "Save as New") and removing the pop up that appears after saving "do you want to overwrite or save as a new deck" 

The Overwrite/New Copy dialog is a bit irritating, popping up every time you save. People used to using files are used to having "Save" and "Save As New" as separate functions. 

I also suggest having a Delete Version feature, and a "Move This Version to New Deck" features.

Epix37:
**Save as:** 
this is a good idea!
How exactly should that work though? You would probably want the ability to save it as 
* a fresh v1.0 (no history) 
* one of the 3 default options (1.0, 1.1, 2.0)
right?

So menu wise:
* `SAVE`
 * `1.3 (current)`
 * `1.4`
 * `2.0`
 * `SAVE AS NEW`
    * `1.0`
    * `1.3 (current)`
    * `1.4`
    * `2.0`

How does that sound?

**Delete version**:
In addition to `EDIT > DELETE`, I would just do `EDIT > DELETE SELECTED VERSION (v[version number here])`, which deletes the version currently selected in the dropdown above the deck. 
Or alternatively, expand the delete menu to:
* `EDIT`
 * `...`
 * `DELETE`
    * `COMPLETE DECK`
    * `SELECTED VERSION`
 * `...`

**Move to new deck**:
Similar to above, either have `EDIT > CLONE SELECTED VERSION` in addition to `EDIT > CLONE`? Or expand the menu:
* `EDIT`
 * `...`
 * `CLONE`
    * `COMPLETE DECK`
    * `SELECTED VERSION AS`
        * `1.0`
        * `1.3 (current)`
        * `1.4`
        * `2.0`
 * `...`

I'm not really sure about the submenu of `CLONE > SELECTED VERSION`

GiggliG:
That's a good question about Save As. I just assumed you'd save it as 1.0 of the new deck. Since it's a new deck, you are starting over with that and no history so it probably ought to be 1.0.
Your submenu idea is perfectly workable, but saving as other version numbers may be more complexity than is needed. 

Delete: I like the menu approach, but I don't know about a submenu.
* EDIT
** DELETE (Deck)
** DELETE (Selected Version)
Since there's only two options, having a submenu is overkill.

Clone:
Yeah, with Clone I would just have Clone (entire deck) and Clone (selected version).

Looking good!

MarioZG:
What would be difference between save as.. and clone?

Also for delete version we need to choose what to do with games assigned to it. Also delete? Seems bit a waste of collected data.

Don't get me wrong, but as not too experienced user, i think i would be honestly scared seeing all these options in save menu. Would give me few minutes of thinking what is correct choice for me.

Maybe we just need "set active version" option, to choose which one can be exported to game and assign new games to?

BTW: I think this is very close to crating "git for decks":)

Epix37:
The "set active version" should be already in (64c4536) with the the PRE6 I linked. At least half of it, I think games are currently assigned to the latest, not the selected version (that's an oversight on my part).

The games of deleted versions are indeed a problem since `PlayerDeckVersion == null` gets assigned to 1.0. Not sure how to handle that either. For complete decks I have "default decks", one for each class, to which games of deleted decks can be moved. Maybe something similar... Might have to wait for another release.

There is no real difference between save as and clone at this point. One is accessible via EDIT and the other via SAVE in the edit panel. Having both is just more convinient plus "save as" is more *natural* than clone.

I did go a bit overboard with the options in my previous post :). Removing the last sub menu level for each one, as @GiggliG suggests should make it much more manageable.

MarioZG:
As for delete version, maybe it shouldn't be possible if there are any games assigned to specific version? User can either move games manually and then delete, or keep it for historical reasons (which makes a lot of sense i believe)

GiggliG:
I agree keeping games for historic reasons makes sense, but if you delete the deck it's very hard to get to view those games. Maybe the Stats/Overall frame could do a Class Played filter, to make that easier. 

Something to also consider.

When you save the deck, you still get that "How do you wish to save the deck popup" with Overwrite and Save as New options.
I assume that dialog will be phased out? I hope so anyway, it always made me panic a little (thinking, "If I overwrite, will I erase all my saved deck data?"). Will there still be a need for it?

There's another popup that occurs when you clone a deck, that asks if you want to keep the history. It is very confusingly written.

The popup says, 
Clone Game Stats?
* "Cloned games do not count towards class or overall stats.
* Yes / No

I am never sure which button to press to make a copy that has no stats, and it could be interpreted to mean that if you click Yes, you are agreeing with the text (Cloned games do not count against stats) and if you click No you are disagreeing with that.

This IMO should be rewritten as something like:
Title: History of Games Played
Text: "Do you want to clone the history of games played, as well as the cards? (Whether you click yes or no, cloned games do not count towards overall stats.)"
Buttons: Yes / No

GiggliG:
Will there be a way to move games played from one version to another? I just noticed you have fixed the games not showing up in 1.0 correctly, but the three games I played in version 1.1 are listed under v 1.0. 

Also, unrelated (I'd report this as a general bug, but it might just be this version): When you clone a deck, and keep the game history , the Name column is empty in the new deck's history (and the * which shows there someone conceded is missing too).

MarioZG:
There's bunch o nulll exceptions when you run application without deck selected - https://github.com/MarioZG/Hearthstone-Deck-Tracker/commit/cdcca5f28eb918e656f6cdb295ff271725b0b7eb

Epix37:
I added you as a collaborator. You can now just push things directly to feature/DeckVersioning. 

This way we don't need to to things twice :).

MarioZG:
Thanks. That's pushed now.

Epix37:
- [x]  Added "save as" and "clone selected"
- [x]  Removed the "How do you wish to save the deck" popup
- [x]  Clone dialog now reads:

> **Clone game history?**
> (Cloned games do not count towards class or overall stats.)
> [clone history] [do not clone history]

I think, to get this working for now, moving decks between versions and deleting versions probably has to wait for the next release.

Anything else that's still missing?

Epix37:
Git is still confusing the hell out of me sometimes...

I just wanted to rebase this on master after merging the missing cards feature, but that didn't quite work for some reason.

So I pushed the rebase (this) to master now, everything seems to be working fine. I hope...

Epix37:
I'm going to close this, as it's merged now. We can continue discussion here though.

Epix37:
Latest (master) build: http://www.mediafire.com/download/7ri8h2c0asppz7b/Hearthstone_Deck_Tracker-v0.7.6_PRE7.zip

PR #604, opened 2015-01-25
Deck dust calculator

Herr-Sepp:
! you need to delete config.xml to get the new default value. (or make it manual ExportCard1X,ExportCard2X, ExportCardY)

If you export a Deck you can now see what cards are in your collection missing and how much dust it cost to produce them.
http://i.imgur.com/bFbIFCP.png?1

I try to detect cards with the mana-crystal. Works better for me because of the higher difference to the background and it is on every card the same.

If a Card is exported twice i look at the botton if i can detect the lock (only one Card available).
This can be tricky if the card is glowing "New". i try to move the mouse over the card to remove that glow but i am not sure if this always help.

So please test it.

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -82,7 +82,7 @@ namespace Hearthstone_Deck_Tracker
 		public bool Debug = false;
 
 		[DefaultValue(50)]
-		public int DeckExportDelay = 50;
+		public int DeckExportDelay = 60;
 
 		[DefaultValue(false)]
 		public bool DiscardGameIfIncorrectDeck = false;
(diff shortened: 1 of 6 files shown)


Epix37:
I ran out of "new" cards to test this with, but this might work:

Basically use the same check you use for seeing if the second card exists, but check for the light blue tint instead. Since this *may* be ignored by my "ignore sparkle" shenanigans, checking for the valid pixels below x% of height*width might also work. 

Then, if that it the case do the "move mouse over card" but add a `await Task.Delay(100)` (maybe a little longer) after each movement.

Epix37:
Another idea: save the list of missing cards to the deck, and add a way to display the same message via the menu. (thanks @n4ru :))

I don't mind adding this one myself.

Epix37:
For the config changes: adding this to MainWindow_Load.ConvertLegacyConfig(...) should work.

```
if(configVersion <= new Version(0, 7, 6, 0))
{
	if(Config.Instance.ExportCard1X != 0.04)
	{
		Config.Instance.Reset("ExportCard1X");
		converted = true;
	}
	if(Config.Instance.ExportCard2X != 0.2)
	{
		Config.Instance.Reset("ExportCard2X");
		converted = true;
	}
	if(Config.Instance.ExportCardsY != 0.168)
	{
		Config.Instance.Reset("ExportCardsY");
		converted = true;
	}
}
```

Epix37:
Implemented my 2nd and 3rd comment and refactored all of it to make a bit more clear what's actually happening.

PR #618, opened 2015-01-30
Missing Cards Ingame

Herr-Sepp:
http://i.imgur.com/nmj6Hq8.jpg

If we run Hearthstone in widescreen (16:9 or 16:10...) we can use the blank space on the left to show the missing cards ingame. so it is easy to built them without sending back to windows.

If the option "extra Features" is enabled you can close it with a click on the button. (with lag)
If not the user can close it with a mouse-over.

I show the Missing cards even if the option "Hide in Menu" is set because if you export a Deck you want to see this and you can close it ingame.

### DIFF
--- a/Hearthstone Deck Tracker/MainWindow/MainWindow_Export.cs
+++ b/Hearthstone Deck Tracker/MainWindow/MainWindow_Export.cs
@@ -50,10 +50,64 @@ namespace Hearthstone_Deck_Tracker
 				await controller.CloseAsync();
 
 				if(deck.MissingCards.Any())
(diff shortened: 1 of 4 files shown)


Epix37:
Hmmm. Gerneral idea is good, few things though:

* Mouseover to close doesn't seem to be working for me
* I think this should be displayed in addition to the message on the tracker
* Should probably be an option?

Cu8eR:
@Herr-Sepp - really nice feature - could you improve it a little bit according to @Epix37 ideas? I am looking for this !

Issue #601, opened 2015-01-22
"Highlight cards in hand" doesn't works


WantedThing:
I'm using HearthStone on Europe server with french language.
I tried all different combinations in the settings but i'm still not able to make this work. I searched on youtube for french videos showing your program running and it seems that the problem doesn't concern only me, that's why i write it here.

Cu8eR:
I can confirm it... no highlight

WantedThing:
I realized that it works if you're playing with a deck you have configured and enabled in the HS-Deck-Tracker.

Otherwise, it's buggy, nothing is higlighted except :
-Bonus cards like the piece
-Units or cards that come back to your hand
-Cards in your hand after a disconnection
-Cards in your friend's hand when you go watch him play

I hope it helps^^

Epix37:
So this in only happening in "no deck"-mode, yes?

WantedThing:
Yes.

PR #619, opened 2015-01-31
Update Card.InHandCount for PlayerDrawn deck.

jw013:
A single commit to address #601. Commit message copied below.

Originally, card drawn / play / get methods only updated
card.InHandCount for PlayerDeck, not PlayerDrawn. This commit updates
PlayerDrawn as well to make the "Use No Deck" overlay behavior more
consistent with the overlay behavior for when a pre-configured deck is
displayed. Specifically, cards in hand are now highlighted the same way
whether a pre-configured deck is used or not.


Epix37:
Perfect, thank you.

PR #637, opened 2015-02-04
Archive decks feature

dvide:
Hi Epix,

HDT can sometimes be a little overwhelming when you have amassed a lot of decks. It's nice to clean up your collection every now and again by deleting the old decks you don't want to use anymore. The only problem with doing this is with the history of games that were assigned to the decks you want to delete. You can delete the decks and remove the games along with them, which is unfortunate because that is removing history and statistics, or you can keep the games but they will not be assigned to any decks any more.

This feature allows you to archive a deck instead of deleting it, so that it is hidden out of the way instead. You can still delete a deck if you choose to do so though. It is sort of inspired by how Google Hangouts or Facebook Chat works, if you want to hide old conversations but keep the chat history for future reference.

This adds an item on the Edit menu to archive and unarchive decks. Archived decks will not show up in the deck picker under 'All' or the deck's class. Archived decks will only be listed under 'Archived' until they are unarchived again.

archive
archived

If an archived deck was selected on shutdown it will not be selected again on startup. This was done on purpose to prevent confusion, i.e. if you start the program and by default an archived deck is selected. There would be no immediate indication that the selected deck is archived so I thought it best to prevent this.

I rushed through this implementation so it needs a code review and some testing. I mostly just wanted to suggest this feature and I thought I'd show it with a quick implementation, rather than make the suggestion. The red colour I just picked at random; you can change it to whatever you think is appropriate.

Possible additions could be:
* Disable menus for editing an archived deck, changing its notes, or updating from web, etc. This way the deck must be unarchived first to make any changes. It doesn't seem to do any harm to allow changes though so I personally don't like this.
* If not the above, perhaps if you edit an archived deck it should be automatically unarchived.
* If you play a game on an archived deck, perhaps automatically unarchive it? It would probably be crucial to keep the deck selected though, as the player might go into the next game not realizing that this has happened.
* If you assign a game to an archived deck, perhaps automatically unarchive it?
* If you export a deck to hearthstone, perhaps automatically unarchive it?

What is the 'Undefined' class in the deck picker for? It seemed to me from the code that it would only ever put anything in there if there is a bug. So perhaps removing that entirely would be a good idea, and then we could also repurpose its grey colour for 'Archived'?


Epix37:
This is a very good idea. 

WIth the HearthStats syncing release just around the corner (hopefully by the end of the week), I would want to wait a bit before merging this to make things easier on myself. It does fit perfectly into there though, HearthStats has an archive function, too. 

You pretty much hit the nail on the head with the undefined.

dvide:
Sounds good :) am looking forward to the HearthStats release too.

dvide:
I've updated this to work with the new deck picker. Don't merge this yet as I haven't tested it much. Just wanted to share it with you to see what you think for now.

Epix37:
Awesome! I will check it out tomorrow night!

dvide:
Here's what it looks like with the icon:
untitled

dvide:
I don't think we should sync the archived status with HearthStats because of the way it works on there. It seems that on HearthStats, decks are automatically archived if they aren't used within a week, and there is no way to manually archive or unarchive a deck. I don't like that as I'd prefer to manage it manually, and in any case if we sync that data down from HearthStats people might get confused when their decks disappear.

It also doesn't seem to respect its own rules when games are uploaded via HDT. What I mean is, on HearthStats, I've got decks that have been archived automatically even though I have definitely played games with them in the past week. Perhaps it doesn't automatically unarchive a deck unless the game was added on the site manually or uploaded with their own client. It doesn't seem to respect it when HDT uploads the game (if that makes sense).

dvide:
Just added the feature to automatically unarchive the deck if you edit it. Also 'EDIT > CLONE' creates the cloned deck as unarchived. Because of this, I also changed it so that the cloned deck is automatically selected for you in the same way that saving a new deck is automatically selected. That way it's not creating the clone unarchived but leaving you in the archived class where you can't see it.

Hope that's okay.

dvide:
Question: Do you think auto deck selection should filter out archived decks, or allow it to select them like any other deck? I'm not sure what the best thing would be there. I actually don't use it tbh (always select my deck manually) so I have no real understanding of what it does and its limitations.

Epix37:
- We can achieve this then by just writing "[Archived]" into the note when syncing.

- Sure, the autoselect changes are ok!

- I wouldn't include archived decks in the selection dialog or auto selection. If there's demand for that we can just add an option for it later on.

dvide:
What do you mean about "[Archived]" in the note?

Epix37:
We write "[Archived]" in the note field when uploading decks to hearthstats and parse it to set the archived state again when downloading. Source URLs of imported decks are handled the same way.

dvide:
Oh I see, so that it will sync the archived flag to other computers as well. Yeah, that makes sense, I wasn't thinking of that case.

dvide:
I'm having an issue with deck versions where the master deck has archived=true, but an earlier version has archived=false and it's messing things up. I'm really confused right now :P

Is `DeckList.Instance.ActiveDeck` always supposed to be the master deck, and `DeckList.Instance.ActiveDeckVersion` supposed to be the active version only? Say if you had a deck with 1.0, 1.1, and 1.2, and 1.1 is currently selected, is `ActiveDeck` supposed to always be 1.2 (the master deck which encompasses the others), and `ActiveDeckVersion` is supposed to be 1.1? That's how I thought it was supposed to work, or am I not understanding this?

Epix37:
That's how it should work, yes. 

So, what if you just add
```
foreach(var version in deck.Versions)
	version.Archived = archive;
```
[here](https://github.com/Epix37/Hearthstone-Deck-Tracker/pull/637/files#diff-1f9b45eafda6a5ff5bba7e7df886c9faR103)?

dvide:
Yeah I could do that I suppose. What I'm confused about actually is why `AddUrlToNote` or `PostDeckAsync` only ever seems to run once, with version 1.0 of deck.

I have a deck with two versions, 1.1 and 1.0. When I delete the deck on HearthStats (`HEARTHSTATS > DELETE SELECTED DECK`) and resync, it only seems to run `PostDeckAsync` once for v1.0 (from what I can tell, with a breakpoint). Yet 1.1 is definitely being uploaded as well. Where is it uploading 1.1? :P

Epix37:
[1.0 gets uploaded](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/HearthStats/API/HearthStatsManager.cs#L259), [Versions get uploaded](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/HearthStats/API/HearthStatsManager.cs#L278)

dvide:
Ah okay I see. So I guess the easiest answer is to do what you said, just setting the archived flag on all versions.

Though it seems that there is an issue with it uploading the correct note as well then. If I create a deck and add a note, then add a version 1.1 and then change the note, then I close HDT and delete PlayerDecks.xml. When I reopen it and sync, it will download the old note.

Epix37:
Yea, changing the note does currently not update on hearthstats. I'll change that for the future.

dvide:
Yeah that's another issue. The same is true atm for archiving. If I archive a deck, it won't update the note on HearthStats to add the [HDT-archived] tag because there's nothing that tells it to. But I think that's technically a different issue really that needs to be addressed as well.

Probably shouldn't reupload the whole deck again just for that either, if the API allows you to just update the note itself.

Epix37:
You could call `HearthStatsManager.UpdateDeckAsync(deck, background: true);` after setting the archived state.

dvide:
Is that not a bit overkill? In terms of reuploading the cards and everything all over again. Might be better to create a UpdateNoteAsync specifically for when only the note needs to change.

Epix37:
Currently there's no way to do that, the all fields in the call are mandatory. There's no really a big difference in data anyway. If it turns out to be a problem we can still change that later.

dvide:
Ok fair enough :)

dvide:
Okay take a look at the latest commit (0809e2e). I should now sync archived flag with HearthStats correctly, and it will sync the changes immediately.

I also changed it so that it always creates the correct note from the master deck note instead of the v1.0 note. Did this by adding the note as a parameter to `PostDeckAsync` instead and the caller passes in the master deck's note, [like so](https://github.com/dvide/Hearthstone-Deck-Tracker/blob/0809e2e33411bab6c928745f7622e5e42aa2f424/Hearthstone%20Deck%20Tracker/HearthStats/API/HearthStatsManager.cs#L259-260).

Don't merge this yet as there is still a bit more work to be done and I want to squash these commits together before you merge. Just pusing it now for you to see how it's coming along so far.

Epix37:
Looks good :).

dvide:
Still need to change it so that it won't auto select archived decks, and I want to try out automatically unarchiving the deck if you play a game with it (after a game is successfully assigned and you're back in the menu). I've got an implementation of that already and seems to work okay, but still need to test it and investigate a few things to make sure everything is fine :)

Also I'm thinking the archived 'class' should perhaps hide itself when no archived decks are visible due to the current tags or deck type (constructed/arena).

Epix37:
I made a few (minor) changed to the menubar and added a context menu for the deck picker items. Shouldn't be too much trouble for you to merge, I think.

dvide:
Okay thanks, I'll sort it out :+1:

dvide:
Okay it should be merged nicely now with the new menu items and context menu.

dvide:
I've reworked the `DeckPickerClassItem` selection logic in `NewDeckPicker.Selector_OnSelectionChanged`. This has to do with multiple class selections. This fixes two existing bugs and also allows it to work with the new archived class.

I'll explain how to reproduce the two existing bugs:

1. Click on class like Mage, then Ctrl-click 'All'. Instead of multi-selecting, it replaces the existing selection with 'All'. This was working as intended, because 'All' encompasses Mage anyway. However, this didn't work in reverse. If you click 'All' first, then Ctrl-click on Mage, it allows you to select them both. The changes fix this inconsistency. If 'All' is selected and you try to add a class other than 'Archived' to the existing selection, it will not allow it.

2. Like above, click on a class like Mage, then Ctrl-click 'All'. Now click on 'Mage' again (without Ctrl). Previously it didn't reselect Mage. This was because the underlying control still thought Mage was selected when you Ctrl-clicked on 'All'. It looks unselected because its `OnDeselected` event was raised (by us), but the underlying control still thought that it was selected, so when you click on Mage again it doesn't consider it a new addition to the selection. The changes fix this problem too.

The logic now works with the new Archived class too. 'Archived' can be selected with any other class, including 'All'. All cannot be selected with any class except for 'Archived'. I like it this way because you can select 'All' and 'Archived' together and see all the decks, whether they're archived or not, and the little marker icon on the `NewDeckPickerItem` is really handy for differentiating between them. Same with using 'Archived' with specific classes.

The logic is very nested and fairly hairy but I thoroughly tested all possible code paths and combinations and I'm confident this is working very well now.

Issue #668, opened 2015-02-10
Crashes on start (>v0.9.0), when run without admin permissions


WizardBlack:
Log Name:      Application
Source:        Application Error
Date:          10/02/2015 22:35        :22
Event ID:      1000
Task Category: (100)
Level:         Error
Keywords:      Classic
User:          N/A
Computer:      Win7PC
Description:
Faulting application name: Hearthstone Deck Tracker.exe, version: 0.1.0.0, time stamp: 0x54da54ae
Faulting module name: KERNELBASE.dll, version: 6.1.7601.18409, time stamp: 0x53159a86
Exception code: 0xe0434352
Fault offset: 0x0000c42d
Faulting process id: 0x3204
Faulting application start time: 0x01d04581db03a3ed
Faulting application path: C:\Program Files (x86)\Hearthstone Deck Tracker\Hearthstone Deck Tracker.exe
Faulting module path: C:\Windows\syswow64\KERNELBASE.dll
Report Id: 1913e3bc-b175-11e4-9142-d05099097dda
Event Xml:
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Application Error" />
    <EventID Qualifiers="0">1000</EventID>
    <Level>2</Level>
    <Task>100</Task>
    <Keywords>0x80000000000000</Keywords>
    <TimeCreated SystemTime="2015-02-10T22:35:22.000000000Z" />
    <EventRecordID>23766</EventRecordID>
    <Channel>Application</Channel>
    <Computer>Win7PC</Computer>
    <Security />
  </System>
  <EventData>
    <Data>Hearthstone Deck Tracker.exe</Data>
    <Data>0.1.0.0</Data>
    <Data>54da54ae</Data>
    <Data>KERNELBASE.dll</Data>
    <Data>6.1.7601.18409</Data>
    <Data>53159a86</Data>
    <Data>e0434352</Data>
    <Data>0000c42d</Data>
    <Data>3204</Data>
    <Data>01d04581db03a3ed</Data>
    <Data>C:\Program Files (x86)\Hearthstone Deck Tracker\Hearthstone Deck Tracker.exe</Data>
    <Data>C:\Windows\syswow64\KERNELBASE.dll</Data>
    <Data>1913e3bc-b175-11e4-9142-d05099097dda</Data>
  </EventData>
</Event>

WizardBlack:
[...]

PR #670, opened 2015-02-11
Run Updater.exe as administrator (#668)

dvide:
This will allow the Updater to work if the executable resides in %ProgramFiles%.

This should do what I was talking about in issue #668. This embeds a manifest file into Updater.exe which tells Windows that it needs to run with administrator privileges. When HDT starts Updater.exe, the user should see a UAC dialog prompting them to elevate the process. When elevated, Updater.exe can write to %ProgramFiles%, otherwise it will fail.

http://www.codeproject.com/Articles/17968/Making-Your-Application-UAC-Aware

Tested this by forcing HDT to update (this code is not present here, as it was just for the test). It seemed to do the job. I'm not too familiar with this stuff though so if anyone else has any expertise with this, please take a look and make sure I didn't do anything stupid :)

### DIFF
--- a/Updater/Updater.csproj
+++ b/Updater/Updater.csproj
@@ -31,6 +31,9 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
+  <PropertyGroup>
+    <ApplicationManifest>app.manifest</ApplicationManifest>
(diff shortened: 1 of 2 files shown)


dvide:
manifest

Here's how I added it. Was pretty simple, but just so you know for future maintenance.

PR #672, opened 2015-02-11
Save file dialog on EXPORT > SCREENSHOT and SAVE AS XML

dvide:
This changes `EXPORT > SCREENSHOT` and `EXPORT > SAVE AS XML` to present the user with a save file dialog, so they can choose where to save the files. This seems more natural and also allows the feature to work when the executable resides in %ProgramFiles% (issue #668).


dvide:
Also changed the 'Save to file' button on the Log tab to use the SaveFileDialog too.

PR #673, opened 2015-02-11
Fix updater updating itself

dvide:
This allows the updater to update itself. 

Previously wasn't copying the Updater_new.exe to the correct path.


dvide:
Unfortunately means users will have to download the latest version manually in order to even get this :)

Epix37:
Oh wow.. uhm.. yea.. 

I think the solution here is to just rename the Updater.exe (and make according changed in the updater and the tracker).

dvide:
That's actually a good idea, renaming the exe. Didn't think of doing that.

PR #721, opened 2015-02-24
Added Win-Loss count to NewDeckPicker

Coywolf:
I wanted to see actual win-loss counts for my decks, particularly for arena.

Textblock for WinPercentString was copied into a Viewbox to allow
scaling, so large numbers don't get clipped. Both Textblocks were moved
to a grid for positioning.


Epix37:
I like it, thank you.

PR #727, opened 2015-02-26
Added Fatigue counter that shows up once there are no cards in a deck.

mnixo:
http://i.imgur.com/bFPI4wA.png

Hi there!
I tried to add this functionality since some times it is hard to keep track of the fatigue damage since the history on the left is so short.
The fatigue counter is located under the the card counter of each deck but only is displayed once the deck has 0 cards.
But I ran into a little problem. I'm not sure why, but I'm unable to set this counter as separate option (it just won't show up) so I ended up hooking it up to the card counter, meaning that whenever the card counter is enabled so is the fatigue counter. It makes sense, altough it is not correct in terms of implementation. I'm hoping once it gets merged someone can quickfix it or if someone could share some light in what I might be missing, I'd be grateful. 
Hope this is useful!

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -192,6 +192,9 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue(false)]
 		public bool HideOpponentCardCount = false;
 
+        [DefaultValue(false)]
+        public bool HideOpponentFatigueCount = false;
+
 		[DefaultValue(false)]
 		public bool HideOpponentCardMarks = false;
 
@@ -209,6 +212,9 @@ namespace Hearthstone_Deck_Tracker
 
 		[DefaultValue(false)]
 		public bool HidePlayerCardCount = false;
+
+        [DefaultValue(false)]
+        public bool HidePlayerFatigueCount = false;
 
 		[DefaultValue(false)]
 		public bool HidePlayerCards = false;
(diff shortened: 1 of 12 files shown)


Epix37:
Hey, that looks good. 

What exactly would you want to do different from your second commit? Have the counter as a separate (moveable) object on the overlay?

mnixo:
Thanks!
Yeah, I thought it would make sense to make as a distinct option from the card counter but was unable to do it. As it is right now, it is being called in it's own option element, which I can't make it work (it doens't show up by itself), and in the card counter option element. What's left to do is to decide what method is the best.

Epix37:
I think this should be fine. 

Maybe just change 
``` 
LblPlayerFatigue.Visibility = Config.Instance.HidePlayerFatigueCount ? Visibility.Collapsed : Visibility.Visible;
```
to 
``` 
LblPlayerFatigue.Visibility = Config.Instance.HidePlayerFatigueCount || Game.PlayerFatigueCount == 0 ? Visibility.Collapsed : Visibility.Visible;
```
(and for opponent respectively) in the overlay and windows?

Epix37:
Never mind that, it's already set to an empty string.
Yea, I think this should be good.

Epix37:
Oooh. I'm a bit slow today apparently. I finally understood the issue.

Old config values are not overwritten by default. 
To make that happen add this:
```
if(configVersion <= new Version(0, 9, 6, 0))
{
	if(!Config.Instance.PanelOrderPlayer.Contains("Fatigue Counter"))
		Config.Instance.Reset("PanelOrderPlayer");
	if(!Config.Instance.PanelOrderOpponent.Contains("Fatigue Counter"))
		Config.Instance.Reset("PanelOrderOpponent");
	converted = true;
}
```
to `MainWindow_Loaded.ConvertLegacyConfig()`

mnixo:
I knew I was missing some sort of small detail. Awesome.
Everything is in the right place now. I've also done some tests and everything seemed smooth. Ready to merge.

Epix37:
And done! Thank you!

PR #744, opened 2015-03-04
Don't put URL in note on import or sync

dvide:
Since the introduction of the 'open website' button, I assume we don't want to put the URL in the note anymore when importing from the web or NetDeck, or when we download decks from HearthStats.

I wanted to build on this commit in the archived decks branch, but it doesn't seem logically related to that feature so I created a separate pull request for it. This way the work is not going to be sitting in that branch for however long it's going to take to finish that feature off and be merged into master.

I also changed the tag we put into the HearthStats note to `[HDT-source=<url>]` rather than `[source=<url>]`. This is mainly because I want to use `[HDT-archived]` instead of `[archived]`, because it's more likely to not be used by the user (small chance but possible). So I thought to be consistent I'd change the URL tag too.

It still reads the `[source]` tag correctly though, since I've made the 'HDT-' prefix optional in the regex. This means it will still download decks correctly that users have previously synced to HearthStats, but new uploads will use `HDT-source` tag instead. It shouldn't result in any changes to user experience; it will simply be more consistent with the `[HDT-archived]` tag once that is added. If you don't like this whole thing though, I don't mind removing it and just sticking with `[source]` since it's not too important.

### DIFF
--- a/Hearthstone Deck Tracker/HearthStats/API/HearthStatsAPI.cs
+++ b/Hearthstone Deck Tracker/HearthStats/API/HearthStatsAPI.cs
@@ -708,13 +708,8 @@ namespace Hearthstone_Deck_Tracker.HearthStats.API
 		{
 			var note = deck.Note;
 			if(!string.IsNullOrEmpty(deck.Url))
-			{
-				var urlString = "[source=" + deck.Url + "]";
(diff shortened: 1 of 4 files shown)


Epix37:
Oh, I meant to change that. That you. I'll merge this when I get home.

PR #746, opened 2015-03-04
Fix notes and stats context at startup

dvide:
Notes and stats context wasn't set to the active deck at startup. This could lead to a null ref exception if you try to edit a note immediately after starting up.

### DIFF
--- a/Hearthstone Deck Tracker/MainWindow/MainWindow.xaml.cs
+++ b/Hearthstone Deck Tracker/MainWindow/MainWindow.xaml.cs
@@ -331,7 +331,10 @@ namespace Hearthstone_Deck_Tracker
 			//UseDeck(lastDeck);
 			//}
 			if(DeckList.Instance.ActiveDeck != null)
+			{
+				SetupNotesAndStats(DeckList.Instance.ActiveDeck);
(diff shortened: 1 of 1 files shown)


dvide:
Not sure whether this same bug happens in other circumstances, like auto deck selection or other cases. That needs to be investigated. I say this because it looked like it was only setting the notes context on the deck picker selection changed event, so I added it to startup as well, but it may also need to be added to other circumstances where the deck selection changes that I didn't look into yet.

Epix37:
OnSelectedDeckChanged is called on start if a deck is selected, and otherwise the edit menu is disabled, so you can't access the note dialog in that case. Stats only shows the overall part if no deck is selected. 

This should not be an issue / the changes don't have any effect? Or am I missing something?

dvide:
To reproduce on master, open HDT (with a deck active) and immediately go to `EDIT > NOTES` and start typing a note. Don't click on anything else. It should crash. Also stats window won't have the title of the active deck like it's supposed to.

`DeckPickerList_OnSelectedDeckChanged` isn't called on start for me. It seems that `ListViewDecks_OnSelectionChanged` is called, but it doesn't raise the `OnSelectedDeckChanged` event on start because `_reselecting` is true. Is that the real bug here?

PR #750, opened 2015-03-05
Use correct version's note when syncing

### LINKED ISSUE #745: Adding a note, doesn't sync the deck on hearthstats
Have a deck without notes
I add a note
I try sync now and force sync
But the note is not updated on the hearthstats deck :(

Regards,
Jean

dvide:
Thought I'd cherry pick this out of the archived branch so it can be used when fixing #745.

Note that it doesn't fix #745 by itself. It just takes the note from the latest version (the master deck) rather than v1.0 of a deck.

### DIFF
--- a/Hearthstone Deck Tracker/HearthStats/API/HearthStatsAPI.cs
+++ b/Hearthstone Deck Tracker/HearthStats/API/HearthStatsAPI.cs
@@ -192,7 +192,7 @@ namespace Hearthstone_Deck_Tracker.HearthStats.API
 			}
 		}
 
-		public static async Task<PostResult> PostDeckAsync(Deck deck)
+		public static async Task<PostResult> PostDeckAsync(Deck deck, string notes)
 		{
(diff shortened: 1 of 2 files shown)


Epix37:
Why not call `AddSpecialTagsToNote` in  `PostDeckAsync` as well? The extra `notes` parameter seems confusing.

dvide:
That's the way it used to work, but when `PostDeckAsync` is called from `UploadDeckAsync` it passes in v1.0 of the deck. This deck's note won't necessarily be the same as the latest note. For example, imagine you don't login to HearthStats and you create a new deck. Then you add a note to it. Then you edit the deck to v1.1. Then you change the note. Then you log in to HearthStats and sync. It used to upload the old note from v1.0 and not the latest note, because v1.0 of the deck is passed to `PostDeckAsync`.

Epix37:
Makes sense.

PR #760, opened 2015-03-07
small features added: record opponent rank and breakdown by type cards left in deck 

MarioZG:
3 changes:
 - on gameplay overlay, added breakdown of left cards by mechanic
 - added opponent rank to note screen. I had to change when note form is displayed to achieve that - you have to leave match result screen before you get your rank and game type.
 - in deck list display win% for total and selected version (i just have seen you added option related to this display, so this might not be relevant anymore)

### DIFF
--- a/Hearthstone Deck Tracker.sln
+++ b/Hearthstone Deck Tracker.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Hearthstone Deck Tracker", "Hearthstone Deck Tracker\Hearthstone Deck Tracker.csproj", "{E63A3F1C-E662-4E62-BE43-AF27CB9E953D}"
 EndProject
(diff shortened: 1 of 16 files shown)


Epix37:
Hmmm...


1) I feel like this doesn't belong in the overlay. It clutters the screen *a lot* and is probably not really all that useful during a game. *If* you are at a level where that information is useufl for you, you most likely know all cards by name anyway and can see the relevant information by glancing at the cards left once. I could see that in some "deck detail" screen however, with pie charts and stuff like that.

2) Opponent rank *can* be determined the same way the player rank currently is. Two rank assets are unloaded after the game (or one if they are the same?). In addition, the player rank also get's unloaded at the start of the game (the medal in the menu). From that it should be terminable what the opponent rank was. So *if* this get's added, I would prefer it to be automatic. Also it would need to be added to the notefield (like [HDT-opprank xx]) of the game when uploading to hearthstats, since they currently don't store the opponent rank).

3) Yea, should no longer be necessary.

MarioZG:
1) I found is useful in few decks where i need too know how many minion types and spells was left in deck. Just being lazy and cannot be asked to count :)  It sits below card list, space is not used anyway. Could do as option?

2) I'll have another look - couldn't find it in logs :/

Epix37:
1) For me it sits outside of the screen for all but one row :). It's too much to just slap it on the end of the deck. I would still vote for the "deck detail" flyout, at least as a base. Otherwise it feels like we skipped a step. We can possibly extend parts of it to the overlay from there, but in a more "organized"(?) manner. As a separate, moveable and scalable panel, etc.

2) it's [this stuff](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/HsLogReader.cs#L435).

PR #848, opened 2015-04-01
Changed context menu. now has checkmarks when the options are activated.

Thristle:
changes from the context menu affect the main window menu and the other way around


Epix37:
Looks good, thanks!

PR #872, opened 2015-04-05
Added more types of curves (hp and attack)

Thristle:


Epix37:
All in all pretty good idea an execution.

Few things though:
1) "NEW DECK" should stay it's own menu item. In general moving it into deck might work, but there's not really any need to do so. Plus when no deck is selected, the "DECK" item gets disabled, making it impossible to add decks when do you don't have any. This could obviously be changed but I would prefer the former.
2) All menuitems, etc. should be all upper case
3) "CURVE" does need to have it's own menu item (on that level). I see 3 options: 
3.1) Moving it into options somewhere (not very accessible)
3.2) Moving it into DECK (not sure)
3.3) Adding a context menu to the manacurve that lets you pick. I like this one the best by far.
In addition the manacurve would need some indicator of what's currently selected.

Epix37:
3.4) (I like this even more) using the same, or a similar, selection the deckpicker uses:

(mockup)
screenshot 2015-04-05 16 46 02

Thristle:
1)i can put the new deck button back but then the menubar is too short and i didn't want to change sizes 
2)i'm going with ur mockup i just hope i won't screw the whole UI up

Epix37:
2) you can basically just copy-paste from `NewDeckPIcker.xaml` :). The manacurve in the UI will end up being a bit shorter but that's probably fine.

Epix37:
I also wouldn't mind helping out. Just let me know.

Thristle:
should be 99% done. you might wanna change colors/sizes.
the code is not dynamic as i wanted it to be (most annoying- the color changes), I need to find an evaluation command (enter code as string) and than i can make it dynamic

Cu8eR:
is it something similar to https://github.com/Epix37/Hearthstone-Deck-Tracker/pull/324 ?

Epix37:
More or less. Rather than adding new graphs it loads the health / attack values in the current mana curve.

Epix37:
I changed it a bit to fit the general style better:
screenshot 2015-04-05 23 13 59
(and implemented it in a slightly different way)

Cu8eR:
OK, but it is a little bit confusing with mana crystals for HP and attack :( is there any idea how to make it better?

I would suggest:
- add a small icon (mana, hp, attack) before the text "mana"/"health"/"attack"
- hide mana crystals for health/attack tab

Epix37:
screenshot 2015-04-05 23 26 42
screenshot 2015-04-05 23 27 15
screenshot 2015-04-05 23 28 03

I certainly don't disagree but none of those look any good in my opinion. There's just too much space.

I don't think the icon on the "tab" is necessary, neither does it fit the theme. 
The other option I see is removing the graphics all together and going with something more "metro-y"

Cu8eR:
I disagree. It looks really good with attack and heath icons... just add mana icosn for mana tab and it would be great

Epix37:
The numbers on the attack icons are way off-center and I feel like you have to look twice to even realize that those are health icons in the first one. They just look out of place. Not happening :). At least not with those.

I'll try metro style (gonna post that tomorrow) and/or sleep on just leaving it as mana icons. From the 4 options above it looks by far the best. And since you can clearly see which one is selected it shouldn't be too confusing anyway.

Cu8eR:
100 people, 100 feelings :) It looks good for me, yes, there is a small problem with centering but better than mana crystals. Dont forget that there is a same problem in game in cards - attack and hp is not centered on the card. I vote for health/attack icons

Epix37:
This would be the "clean" alternative:
screenshot 2015-04-06 00 05 55

Idk, the longer I look at it the more I like it. Fits the new selection above I guess.

PR #885, opened 2015-04-10
Add Misha, Huffer and Leokk

bmichotte:
On [HSTracker](https://github.com/bmichotte/HSTracker), when you play Gang Up, I copy the cards into the deck (either the player's one or the opponent's one). 

If you Gang Up a companion animal, you will receive Misha, Huffer or Leokk, and these cards were missing. I don't know if you track these cards the same way I do, but it's my time to give you some images ;)


Epix37:
I'm slightly shocked I didn't have those already, lol. Thanks :).

culinko:
I don't want to make another issue, so I report it here. I found out that grove tender has also images "gift of mana" and "gift of card" and anodized robo cub has images "attack mode" and "tank mode" (which are slightly different btw.). If those cards have images for all of their "choose one" choices, should the other druid cards have 3 images as well? (e.g. Cenarius or Keeper of the Grove).

PR #903, opened 2015-04-19
Fix card lock detection

condone:
The CardHasLock method was always returning false due to maxBrightness being incorrectly set to 0.


Epix37:
Thanks :).

PR #956, opened 2015-05-12
Display win/loss for all version of deck in the left side

disshishkov:
Usually changes between versions of a deck are small and option to see win/loss for all versions the deck near win/loss for current version seems is nice.
hs_total_stats


Epix37:
You can set what stats to display (all, selected version, selected major, ...) in `options > tracker > stats > display`, making this redundant if "all" is selected.

disshishkov:
hm, yes, if will be chosen `all` the left stats will be useless, but with other options are helpful, because I can see and all and current version in one time.

Does you new plugin system allow to do the same that I can wrap up it into plugin?

Epix37:
*Might* be somehow possible with a plugin, but I wouldn't know how.

A plugin that switches Config.DisplayedStats while a key is pressed could work.

Issue #473, opened 2014-12-19
Feature Request: gradient removal


jsogvndsi:
This is completely cosmetic but the list of decks having a gradient coloured background clashes jarringly with the smooth Windows 8 theme the rest of the application has.  A simple removal of the gradient and using a solid, soft colour for each would hugely improve the look of the application.

Epix37:
See my post in [#979](https://github.com/Epix37/Hearthstone-Deck-Tracker/issues/979#issuecomment-109247537) (work in progress).

Issue #979, opened 2015-05-18
Deck Picker: Separating select and use.


Epix37:
To make features like #977 and many others work/easier, selecting a deck has to no longer put the deck in "active" mode. 
Two main reasons: 
1) Selecting a deck reloads the log
2) Not possible to select another deck while "auto select deck" is on during a game

Possibly something like this: http://imgur.com/a/nNZ9t
(Use button on all decks, only on selected left/right)

I've been wanting to do this for a long time now, but always put if off because I can't think of a good way to do it.
* Do you like any of the button placements? What/how would you change it? *(I'm not entirely happy with any of those yet)*
  * An alternative would maybe be to place a universal "use" button somewhere. 
* Should there be a global indicator of what deck is currently set active? The decklist on the right will not always represent the active deck now. *(I tried placing a deck-item outside of the picker to indicate the active one, didn't really look any good)*

culinko:
I think a double click or right click -> use deck would work for me. The used deck could be in bold font and selected deck could have the border around it (like in 2nd picture). But I'm not the typical basic user, so maybe there should be a way to distinguish it a little bit more.

Also, if you are in a spectator mode (even when automatic deck picker is off) and you are switching up decks, the tracker is really unresponsive and it could even freeze. Sometimes I wanted to look at my decks during a spectator game, but it's not really pleasant experience. It makes sense now, because you said it reloads the log. So this would fix this issue as well.

GiggliG:
It's an interesting problem. 

For showing the active deck, maybe you could change the border, or put a little symbol in the name box, like you do with the (N) for Notes. I think a red REC circle like on tv remotes would work pretty well. Placement: maybe bottom of 

I think you need a "Go to active deck" function too, maybe in rightclick

You could put a little horizontal function bar just below the All - Arena - Constructed headings, with buttons for functions like:
* Use current deck
* Go to active deck
* Toggle sort by class cards first
* compare deck (select one deck, click compare deck button, click second deck, and it compares them)
* Clear tag filter
* View Tags selection (I always thought this should be more visible)

Just some ideas.

Elitesparkle:
> Do you like any of the button placements? What/how would you change it? (I'm not entirely happy with any of those yet)

Placing the button there doesn't look nice.
I would opt for a mouse listener (e.g. double click) instead.

> An alternative would maybe be to place a universal "use" button somewhere.

You could add a new option called Use (or Active) in the Deck menù.

[...]

PR #981, opened 2015-05-20
Minor GUI enchancement-Changed the look of decks in deck list

Paradowski:
This update is inspired by #473 :)

I've made some changes to the list of created decks:
+ As you can see, there is much more easier now to notice currently chosen deck(opacity of unchecked decks is set to 60%).
+ The background gradient of each deck has been changed - in my opinion it looks now way better and modern than the previous gradient (but it's only my opinion ;) ).

image


disshishkov:
don't like this gradient

Epix37:
See my post in [#979](https://github.com/Epix37/Hearthstone-Deck-Tracker/issues/979#issuecomment-109247537) (work in progress).

PR #1140, opened 2015-07-02
Export deck screenshots to imgur.com

andburn:
Added an option under `Tracker > Exporting` to allow deck screenshots to be uploaded to imgur.com. Consists of new `Imgur.cs` class for uploading. Addition of config values for API key and enabling the feature. Changes to `MainWindow_Export.cs` and `MessageDialogs.cs` to cater for uploading.

Also, fixed small issue with screenshot file name being incorrect when selected deck is not active. It turns out this will also fix a crash when no deck is active and screenshot is pressed #1152.

@Epix37 I added the API key I used when testing `public string ImgurClientId = "7e1ab33fda207c2"` to `Config.cs`. If you add it to a release you will probably want to register your own anonymous key for HDT at [imgur](https://api.imgur.com/oauth2/addclient) - takes like two minutes.

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -592,6 +592,12 @@ namespace Hearthstone_Deck_Tracker
 		public bool ShowExportingDialog = true;
 
 		[DefaultValue(false)]
+		public bool DeckScreenshotToImgur = false;
+
+		[DefaultValue("7e1ab33fda207c2")]
+		public string ImgurClientId = "7e1ab33fda207c2";
+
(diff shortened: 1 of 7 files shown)


FitzcairnHS:
Will this export to imgur option will be added soon ?

Epix37:
@FitzcairnHS Yes, this will be in the next update!

@andburn Awesome, thank you!

I removed the option and instead replaced it with a dialog that provides the options to "save only", "save & upload" and "upload only". That might be a bit more handy and avoids yet another hidden feature :).

andburn:
Cool, the dialog makes much more sense.

Issue #1162, opened 2015-07-14
Activating Import Selects 0 Mana for any import.


theSLiKbandit:
Attempting to import a deck into Hearthstone starts off by selecting the 0 mana filter, then attempts to fill out the deck without ever switching back between mana filters.

andburn:
Ok, I'll see if I can figure it out, can you give me the following info: 

1. how often does it happen, everytime or just sometimes
- are any of the mana crystals selected before the import
- the resolution you are running Hearthstone at
- and is it fullscreen or windowed
- in the tracker options what does it say for the *Speed* value in `Options > Tracker > Exporting` (e.g. *"Fast (40ms)"*)

theSLiKbandit:
1. Sometimes
2. None
3. Custom resolution is used for the Hearthstone window.  I play windowed while the monitor is 1366 x 768.
4. Speed value is currently as 40 ms.

andburn:
Thanks. It might be something to do with the windowed setup. I'll test it out some more and see.
You say your monitor is 1366x768, but what do you have the resolution set to in the Hearthstone options, I suppose it has to be 1024x768?

theSLiKbandit:
The resolution displayed hearthstone options is blank.  I have the window custom fit so that it does not extend into the taskbar.

andburn:
Ok, sounds like the custom window size is the problem. I'll try and replicate it, and see if I can fix it.

PR #1214, opened 2015-08-13
Make export detection regions relative to window height

andburn:
Changed the exporting "rectangles" used for detecting the mana crystal, card lock etc. and made them relative to the window height. It should make the detection more accurate at lower resolutions, and be a possible solution to #1162.

Also, for OCD reasons changed the `DeckTest` to cater for the additional parameter in the `Card` constructor :smile:.


Epix37:
Makes sense, thanks. 

Uh.. yea.. I meant to fix the tests :X.

PR #1263, opened 2015-08-19
Implemented new art for card chances.

zachwlewis:
Added themed frames for draw and holding chances.

hdt-chance-frame


Cu8eR:
Nice, really nice!

zachwlewis:
Also updated the card counter. I'm still working on getting these new assets into the individual windows, so don't merge yet.

hdt-skinned-card-frames

zachwlewis:
Alrighty. I slimmed down the frames and I added them to the Player and Opponent windows. I also increased the width of those windows slightly to give the elements room to breathe.

We should be good to pull! :shipit: 

hdt-slim-card-frames

Epix37:
10/10 !

PR #1293, opened 2015-08-21
Fix Dreadscale not being added to ActualCards

### LINKED ISSUE #1291: Dreadscale missing in deck editor
Dreadscale seems to be missing in the deck editor. As far as I'm aware I'm using the latest update. There is an image for him in the images folder and cardDB.enUS.xml contains data for him (the .json file doesn't but that one also doesn't include Acidmaw). It doesn't show up in the list or when searching. I've tried both English and German to see if changes the language has any effect but it didn't in this case. I've check the other classes and it also doesn't show up there.

capture

Acidmaw does show up, and I haven't noticed any other TGT cards missing.

capture

andburn:
Dreadscale's id is *AT_063t*, the *t* stops it from being included in *ActualCards*.
Couldn't think of a better way than adding an id check at the relevant points, its not great but it works!
Fixes #1291

### DIFF
--- a/HDTTests/Hearthstone/CardDbTest.cs
+++ b/HDTTests/Hearthstone/CardDbTest.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿using System.Linq;
+using System.IO;
 using Hearthstone_Deck_Tracker.Hearthstone;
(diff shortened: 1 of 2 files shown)


Epix37:
Why blizzard why,,,

PR #1298, opened 2015-08-22
Roughed in new card marker UI.

zachwlewis:
I added a new control to display pretty card markers. I think everything is working properly, but it definitely needs testing and proper positioning (as in the picture below).

hdt-card-marks


MechaLynx:
This is super-sweet but I have a couple of worries to voice: The icons blend in with the game too much, or at least it seems like it. The indicators we have right now stand out clearly and so do the numbers in your version, but the icons might be too hard to distinguish. It might be really bad, considering the variety of card backs we have available.

Also, I have no idea, just by looking, what the two symbols on the rightmost cards are - I can deduce they're returned and spell-created cards, but at least the spell icon should be more clear. There is a sort of contradiction, from a UX standpoint, in using letters for some indicators and icons for others. You already solve the problem of having the turn marker replaced with the letter by combining them properly, but I think either icons or letters should be picked and stuck to.

All this said, I believe this __should go into the main branch__, if only for the combination of turn indicators and type markers on the cards. The issues I'm talking about can be fixed later and shouldn't be the source of any problems for users, apart from the _possible_ problems I outline above.

---

On the more constructive side, here's what I think the icons _should_ be for more clarity, unless letters are chosen (in which case they should all have clearly contrasting colors from their backgrounds and the cards):

  - `keep` should have no icon (what it is should be clear from contrast with `mulligan`)

  - `mulligan` should have an icon closer to `arrow_switch` from the [silk icon set](http://www.famfamfam.com/lab/icons/silk/previews/index_abc.png) although not necessarily identical of course

  - `returned` should be something like the [recycling symbol](https://en.wikipedia.org/wiki/Recycling#/media/File:Recycling_symbol.svg)

  - `spell` can work through contrast with `returned` if left with the present icon.

  - `coin` is 100% clear, no changes needed

I'll try to put up a canvas-based tester to allow some comparisons on icon contrast based on the card back so that it can be clearer that the icons don't unintentionally blend in with colors from the card backs. Alternatively of course, they could be placed about 90 degrees clockwise to their current position to allow more distance from the card back and have a slight border (like the letters and coin icons have) to contrast with the board background (although in that part, it never changes).

zachwlewis:
Yeah, the keep and mulligan icons were just something I put in quick to see how the implementation would look. The icons actually sit pretty snugly on the background frame, so I don't really think there will be any issues with visibility.

Epix37:
Any immediate changes coming or shall I merge?

zachwlewis:
This is a fine place to merge. You'll probably want to update the positioning code to place them better, but I think it all works.

Issue #1316, opened 2015-08-25
Cloning a "Selected Version" of a deck results in the most recent version being cloned instead.


Contestest:
For example, if I have a deck with v1.0 and v1.1 saved in it. Trying to clone only v1.0 of said deck results in v1.1 being cloned instead.

Tested with a deck with v1.0-v1.4 saved, regardless of the selected version, v1.4 is always cloned

Hearthstone Deck Track Version 0.10.13

Epix37:
fixed with v0.10.14

PR #1318, opened 2015-08-25
Fixed bug #1316 and some formatting

MechaLynx:
Made some formatting changes to this pair of methods for clarity.

When cloning a selected version, it will use the selected version as the current deck to clone, rather than cloning the entire deck and then resetting versions (which resulted in bug #1316).

Line 221 is the only real change, the rest is just cleanup.

### DIFF
--- a/Hearthstone Deck Tracker/MainWindow/MainWindow_Edit.cs
+++ b/Hearthstone Deck Tracker/MainWindow/MainWindow_Edit.cs
@@ -169,6 +169,10 @@ namespace Hearthstone_Deck_Tracker
 
 		internal async void BtnCloneDeck_Click(object sender, RoutedEventArgs e)
 		{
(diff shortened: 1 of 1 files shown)


Hearthstone Deck Tracker/MainWindow/MainWindow_Edit.cs
```diff
@@ -213,6 +218,8 @@ public void ArchiveDeck(Deck deck, bool archive, bool saveAndUpdate = true)
 		internal async void BtnCloneSelectedVersion_Click(object sender, RoutedEventArgs e)
 		{
 			var deck = DeckPickerList.SelectedDecks.FirstOrDefault();
+            		deck = deck.GetSelectedDeckVersion();
```
Epix37: I see a nullreference exception waiting to happen here. Should probably be moved two lines down.

PR #1402, opened 2015-09-06
Add last played date for decks

ko-vasilev:
I added deck's last played date in list of decks.
It is under Use button for Layout1, and at the same position for Layout2 (under stats).


Epix37:
Maybe as an option? Not a huge fan of this being on by default. 
* You can sort by last played already
* Additional clutter
* `USE` button is "harder to hit"

Alternatively, what might work is if we add a tooltip to decks (or an additional panel somewhere). Containing some more information on statistics as well as this. 

What do you think?

ko-vasilev:
How do I sort by last played? Did not know I can do this.
Can definitely make it as optional thing to display.

About "Use" button I think it is still big enough.

Just personally I'm not playing too often to remember which decks are "more recent" and seeing this label easilly will help me a lot :)

Epix37:
`MAIN > SORT / FILTER`

ko-vasilev:
I still think it will be a nice feature to have. Do you want me to add an item to Options menu or you will do it yourself? Or you don't like this feature that much?

Epix37:
Nah, it's fine. Yea, if you could implement the option, would be great :).

Cu8eR:
Any progress?

Issue #1456, opened 2015-09-12
Gold progress bug (minor)


HugoKornelis:
I am glad to hear that the previous issue where the gold tracker had stopped working has been fixed, I have now re-enabled the tracker, and it works nice in Play mode,
However, during an Arena run the gold tracker apparently thinks I should be rewarded too. After having won a few arena fights, it is now telling me that some wins have not been awarded, and that I have "7/3" wins. As far as I know, arena games are not counted for the three-wins-ten-gold reward, so HDT should not expect to see a gold reward.

GaryKing:
Yep I'm seeing this too. It seems to think Arena games should reward the 10 gold for every 3 wins, when it shouldn't.

chucklu:
arena games are not counted for the three-wins-ten-gold reward
I do not know this rule
I will try to fix this later

chucklu:
else if ((DateTime.Now - gameState.LastGameStart) > TimeSpan.FromSeconds(10)
                     && game.CurrentGameMode != GameMode.Spectator)
            {
                GoldTracking(logLine, game);
            }
this code just except  Spectator game mode,so we should except arena mode too?

chucklu:
public enum GameMode
	{
		All, //for filtering @ deck stats
		Ranked,
		Casual,
		Arena,
		Brawl,
		Friendly,
		Practice,
		Spectator,
		None
	}
we have 8 game mode in the decker
@HugoKornelis , Spectator and Arena should not include in gold tracker.
so, what abour the other modes?

riQQ:
Only Brawl, Casual and Ranked games count towards the 10 gold per 3 wins bonus.

chucklu:
@riQQ Thank you!I will fix this and send a new pull request to @Epix37

LizardDude:
Arena matches still count toward the gold tracker.

Issue #1278, opened 2015-08-20
Gold Progress is not tracking


InsidiousJFP:
After the last update HDT stopped tracking gold progress wins. Tried restarting HDT and Hearthstone but HDT still reads "Wins: 0/3 (10/100G)" even when I log in the next day it still reads the same. I really like this application and thanks a bunch for your hard work. :)

Epix37:
That's actually a bug in hearthstone at the moment as far as I know. Currently there is no way to make this work again.

InsidiousJFP:
Bummerino....oh well, thanks for the quick response.

thufirh:
I was looking through the game logs trying to see what I could find, and after gaining the 3 wins gold I noticed the following in the Rachelle logs:
D 10:05:35.2196986 RewardUtils.GetViewableRewards() - processing reward [GoldRewardData: Amount=10 Origin=TOURNEY OriginData=0]
D 10:05:39.6329510 VictoryScreen.InitGoldRewardUI(): goldRewardState = INVALID

The second line shows up after every game, and I assume is why you are unable to track progress towards the gold reward. But that first line does show when a reward is given. Could this be used to at least track progress to the 100g daily limit? Also, since the tracker can tell wins from losses you could even make a guess about the progress towards the next pay out (a correct guess assuming the player never plays an un-tracked game.)

chucklu:
Crash report
########## 2015/8/27 20:12:18 ##########
System.IndexOutOfRangeException: 索引超出了数组界限。
   在 Hearthstone_Deck_Tracker.OverlayWindow.UpdateGoldProgress()
   在 Hearthstone_Deck_Tracker.OverlayWindow.Update(Boolean refresh)
   在 Hearthstone_Deck_Tracker.MainWindow.<UpdateOverlayAsync>d__9e.MoveNext()
--- 引发异常的上一位置中堆栈跟踪的末尾 ---
   在 System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c__DisplayClass2.<ThrowAsync>b__3(Object state)
   在 System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   在 System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
[20:11:16] Game: >>>>>>>>>>> Reset <<<<<<<<<<<
[...]

PR #1466, opened 2015-09-13
only Brawl, Casual and Ranked game mode will count gold progress

chucklu:
https://github.com/Epix37/Hearthstone-Deck-Tracker/issues/1456
fix bug reported by this issue

Maybe, we should close this issue https://github.com/Epix37/Hearthstone-Deck-Tracker/issues/1278  at the same time.

### DIFF
--- a/Hearthstone Deck Tracker/LogReader/Handlers/RachelleHandler.cs
+++ b/Hearthstone Deck Tracker/LogReader/Handlers/RachelleHandler.cs
@@ -32,8 +32,7 @@ namespace Hearthstone_Deck_Tracker.LogReader.Handlers
                 else
                     gameState.GameHandler.HandlePossibleConstructedCard(id, false);
             }
-            else if ((DateTime.Now - gameState.LastGameStart) > TimeSpan.FromSeconds(10)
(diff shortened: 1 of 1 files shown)


Epix37:
Does this work reliably? The game mode is often determined after the game has ended.

chucklu:
sorry,I have not tested this fix,I check when the gamemode get the value

chucklu:
I don't have time to test this today,maybe I can do the test later

chucklu:
ok,I have checked this function.
When I won a game in Arena mode,gold track avoid to count gold reward.

Epix37:
Sorry I wasn't quite clear what I mean. The problem isn't that it would track in arena, as game mode in that case is either `Arena` or `None`, but in the other modes (which might be `None` at the time of the check) .

chucklu:
ok,I debug this gold track procedure, actually the game mode is none when we count the gold reward.
we need to count the gold reward after the game mode detected？

Epix37:
I think I fixed it locally now with mode detection. Still need to test later.

chucklu:
ok,you can fix this by yourself.
I was confused by the game mode detection

PR #1505, opened 2015-09-21
secret deduction via game rules

joshtab:
This pull request is NOT ready in it's current state, but I'd like some feedback before I spend more time on it.

This code starts an effort to use process of elimination to deduce available secrets by looking at game events. The only game event I'm hooking into right now is the "opponent attacked" event.

An example:

before
after

Is this a feature you want? I did a little clean up of the secrets code, let me know if this looks like it's going in the right direction.


joshtab:
One thing that confused me: the desired behavior surrounding "stolen" secrets, and why you'd ever need to display of secrets from multiple classes. (e.g. https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/c052f196c1365e305048e668d5cba89b55fff780/Hearthstone%20Deck%20Tracker/Hearthstone/Secrets/OpponentSecrets.cs#L22)

Is it just for Kezan Mystic? Seems like you would know exactly what secret they had.

Epix37:
I would love to see that added, absolutely!

You can get secrets from other classes for example via Nefarian or Spellslinger.

joshtab:
This is ready to be looked at.

Epix37:
That was quick! I will have a closer look later, very excited :).

What I see right now though: I don't think removing the ability to manually grey out secrets by clicking is a good idea. You might either not want to enable the auto-greyout or it could mess it for some reason and this then leaves you with no way to correct it.

Epix37:
I started writing unit tests for this, first thing I noticed is that there is no distinction between a hero attacking and a minion attacking. Freezing trap for example only triggers if it's a minion attacking.

Something I have not tested yet but have no idea what exactly happens:
* Opponent played has Bear Trap and Misdirection.
* Player hero attacks opponent hero - what happens?
I assume the opponent hero is set as `DEFENDING`, Misdirection triggeres and something else is then set as `DEFENDING`. In that case both traps are set as "not possible", but since the Opponent hero was never really attacked, Bear Trap was not actually tested for. (Is this the real behaviour? I am not even sure...)

The same thing should happen with e.g. Freezing and Explosive.

Also, order of play might matter in some cases, I am not sure...

Epix37:
Pushed this to branch [feature-SecretDecuction](https://github.com/Epix37/Hearthstone-Deck-Tracker/tree/feature-SecretDeduction). Please submit any new changes as a PR to this.

What I did so far:
* Implemented HandlePlayerAttack in a way that considers attacker and defender (have not tested this at all yet but should be fine)
* Implemented Test for all "single secret" HandlePlayerAttack scenarios.

Epix37:
For reference:
http://hearthstone.gamepedia.com/Advanced_rulebook#Combat
https://www.youtube.com/watch?v=oHdveuZXoHg
https://www.youtube.com/watch?v=H3d_qlm4Xws&t=6m25s

This is gonna be fun!

dantedog:
No idea what the actual blizzard policy on this is, but before pushing this out, I would make sure that this is ok for a 3rd party app. I can't think of anything deck tracker does that can't be done via pen/paper. This would possibly cross that line. 

That said, this seems like an awesome feature.

Cu8eR:
it is OK :)

joshtab:
Closing this for the other thread https://github.com/Epix37/Hearthstone-Deck-Tracker/pull/1522

PR #1522, opened 2015-09-28
Secret Deduction

joshtab:
This patches some of the edge cases surrounding combat.

When a "fast secret" gets triggered, sometimes it can disrupt combat (freezing trap, for example, will prevent the attack from going through). Any secrets older than the triggered secret need to be narrowed down based on the proposed combat.

On hero->hero combat, misdirection was being hidden even if there were no other targets. There needs to be a minion in play for misdirection to trigger.

### DIFF
--- a/HDTTests/Hearthstone/SecretTests.cs
+++ b/HDTTests/Hearthstone/SecretTests.cs
@@ -120,7 +120,7 @@ namespace HDTTests.Hearthstone
         public void SingleSecret_MinionDied()
         {
             //TODO: this behaviour is not always true. https://www.youtube.com/watch?v=oHdveuZXoHg
-            _gameEventHandler.HandlePlayerMinionDeath();
+            _gameEventHandler.HandleOpponentMinionDeath();
             VerifySecrets(0, HunterSecrets.All);
             VerifySecrets(1, MageSecrets.All, MageSecrets.Duplicate, MageSecrets.Effigy);
             VerifySecrets(2, PaladinSecrets.All, PaladinSecrets.Avenge, PaladinSecrets.Redemption);
--- a/HDTTests/Hearthstone/SecretTests.cs
+++ b/HDTTests/Hearthstone/SecretTests.cs
@@ -20,8 +20,9 @@ namespace HDTTests.Hearthstone
 
         private Entity _heroPlayer,
             _heroOpponent,
-            _minion1,
-            _minion2,
+            _playerMinion1,
+            _opponentMinion1,
+            _opponentMinion2,
             _secretHunter1,
             _secretHunter2,
             _secretMage1,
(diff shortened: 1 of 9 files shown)


culinko:
Also, misdirection doesn't trigger if all of the targets have stealth (with hero->hero attacks). However, if you choose to attack with one of the stealthed minions, it can activate, because your minion lost stealth with the attack.

joshtab:
@culinko Can't repo.

I attacked with heroic strike into opponent. Only one minion in play, stealthed. Misdirection triggered

misdirection

culinko:
Yeah, they changed the misdirection with 3.1. Now it triggers every time there is 1 minion, no matter if it has stealth or not. Blizzard made it easier for you to implement ;)

Epix37:
That's the status on this?

What's working / needs to be fixed?

Cu8eR:
I would like to see this in the next release :) My life in HS would be much easier :)

joshtab:
Only known issue to my knowledge:

Minion deaths when the minion has a deathrattle which puts a new minion in play. Redemption and Effigy will not trigger if there are 7 minions in play after the deathrattle. The code as written assumes these both will trigger.

Not sure how to fix. Could create a list of minions with "deathrattle: add minion to play", but sounds annoying to maintain. We listen for the Tag change to identify minion death, but I couldn't identify where the "minion deathrattle about to happen" state is stored. Any ideas?

culinko:
I think if you play effigy before the deathrattle minion, the secret triggers before the minion's deathrattle. If you play it after the minion, the minion's deathrattle will trigger first and effigy won't trigger. I might be wrong, though.

Epix37:
Checking the card text for "Deathrattle: Summon" might work. I think the only cards not matching that would be feugen/stalagg.

joshtab:
Think I'll just list the cards - number of minions summoned matters as well. (e.g. haunted creeper)

Will test @culinko's hypothesis about ordering

joshtab:
For ordering, effigy and redemption behave differently. Redemption is "slower" (it always waits for deathrattles to go first). Effigy happens like culinko said. 

I think this whole feature is in pretty good shape. I know of some edge cases but think the game situations will be quite rare:

* In a single turn, Opponent plays plays mystery secret followed by a deathrattle minion. If Opponent has 7 minions in play and you kill the deathrattle minion on your turn, the secret will not properly be eliminated as effigy. (If the secret was played on a previous turn, it would be properly eliminated).
* Cross class scenarios: Ancestral Spirit + Effigy + Full Board
* Stalagg/Feugen Trigger + Effigy + Full Board
* Miscalculating "Full Board" /w Baron Rivendare + Deathrattles + Effigy 

I'm more worried there's a common scenario I've failed to anticipate (with the hunter combat secrets, for example), but don't know of any problems.

Perhaps it's time to get it in front of people and see how it works?

One visual change I'd like but don't know how to implement: when there are multiple secrets in play, mousing over the individual secret bubbles would change the "possible secrets" overlay to list the possibilities for the individual secret. For example, you might know secret 1 is Counterspell or Ice Block. Your Mage opponent plays a second secret, there's useful information which basically disappears.

culinko:
Another scenarios that can screw things up:
*Sylvanas + Effigy (with Sylvanas played first). Maybe you could check for zone positions at some times? (e.g. if there is a minion in position 7 or deathrattle minion (creeper) on position 6 dying).
*Snake Trap+Noble Sacrifice - not sure how these 2 interact, but if you have 4 minions and 3 snakes spawn, they could block the spawn of Defender. If Noble Sacrifice triggers first, it will be fine, since Snakes can trigger on a full board (same with Bear Trap on full board - if they didn't fix it in 3.1).
*How do you deal with Noble Sacrifice interrupting a face attack? E.g. you have Noble Sacrifice and Bear Trap and opponent attacks your face, but it gets redirected. It should not gray out Bear Trap (I'm sure you have this one covered, mentioning it just in case). Similar with Bear Trap and Misdirection. Or does Bear Trap trigger in any of the scenarios I listed anyway?

Pretty sure there's more interactions that we're missing, but can't think of any right now.

Also I really like the idea of mousing over individual secrets and storing the information separately for each secret. I think hover things are possible to implement, see gold tracking in main menu.

joshtab:
I dealt with the Bear Trap + Misdirection case by looking at a "proposed attack" which is interrupted.  There are Fast Secrets (which can disrupt combat by removing or redirecting the attacker) and Slow Traps (which only get triggered after the combat goes through). Bear Trap is Slow and can  only get gray'ed if the attack goes through. Order of the traps matters for things like Freezing + Explosive, and that's handled as well.

Noble Sacrifice /w Bear Trap should be the same as Misdirection /w Bear Trap, but testing all of the cross-class combos is hard.

Epix37:
> Perhaps it's time to get it in front of people and see how it works?

I plan on releasing v0.11.6 tomorrow. I would say we make this a beta (pre-)release a few days after.

joshtab:
Great, let me know if you need anything

Epix37:
Merged into master.

Epix37:
Pre-release now out: https://github.com/Epix37/Hearthstone-Deck-Tracker/releases/tag/v0.12.3

Epix37:
```
System.ArgumentOutOfRangeException: El índice estaba fuera del intervalo. Debe ser un valor no negativo e inferior al tamaño de la colección.
Nombre del parámetro: index
   en System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   en Hearthstone_Deck_Tracker.OpponentSecrets.SecretRemoved(Int32 id, String cardId)
   en Hearthstone_Deck_Tracker.GameEventHandler.HandleOpponentSecretTrigger(Entity entity, String cardId, Int32 turn, Int32 otherId)
   en Hearthstone_Deck_Tracker.GameEventHandler.Hearthstone_Deck_Tracker.IGameHandler.HandleOpponentSecretTrigger(Entity entity, String cardId, Int32 turn, Int32 otherId)
```
Full log: http://pastebin.com/mv3DiBnE

> un mago tenia 2 secretos y use bengala y el programa se crasheooo, tengo activado para reconocer los secretos automatico

Mage with 2 secrets => flare.


Not sure if this is an issue with the implementation or a special case, the logs contains some messages that should usually not appear: `TMP ENTITY (mokko) NOW HAS A KEY, BUT GAME.ENTITIES DOES NOT CONTAIN THIS KEY`. IIRC, this means there was never an entity id tagchange logged for that name.

PR #1587, opened 2015-10-10
Try to fix Win 8/10 exporting issue

### LINKED ISSUE #1341: issue when exporting
when I attempt to exporta deck, deck tracker chooses the basic card set but never unselects the basic cad set and it results it the deck not being fully exported

quark-zju:
This may fix #1341.

The workaround looks pretty unreasonable but the Win 10 HS search box is
even more mysterious. The current solution is indeed the most direct and
reliable method concluded from hours of trial-and-error.

For more details, see:
https://github.com/ko-vasilev/Hearthstone-Collection-Tracker/issues/25#issuecomment-135025630

I was okay win Win 7 compatibility mode. However, one of my friends was experiencing serious decrease in game 3D performance with compatibility mode enabled. I have tested this patch in my Win 10 laptop.

### DIFF
--- a/Hearthstone Deck Tracker/DeckExporter.cs
+++ b/Hearthstone Deck Tracker/DeckExporter.cs
@@ -221,6 +221,16 @@ namespace Hearthstone_Deck_Tracker
 				SendKeys.SendWait(fixedName);
 			SendKeys.SendWait("{ENTER}");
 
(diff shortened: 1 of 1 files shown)


Epix37:
Pressing enter a second time will open the chat if you previously used it. :/

quark-zju:
Umm... That is a problem. I will try to find another fix when I have time.

PR #1589, opened 2015-10-11
Add icon for board attack, change default position

andburn:
Moved the default icon position to the other side of the board. I left the same positioning scheme so it can still be moved when the overlay is unlocked. As you mentioned, that does leave the side effect that it will move with different widths, but I think its good that it can be moved around.


Epix37:
They can already be moved around freely by unlocking the overlay :).

What aspect ratio are the new values based on?
If they are not, coud you make them 4:3 and modify the [`Canvas.SetLeft`](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/Windows/OverlayWindow.xaml.cs#L614) to use 
```C#
DeckExporter.GetXPos(Config.Instance.AttackIconPlayerHorizontalPosition / 100, (int)Width, ratio)
``` 
 (and the opp respectively)?

Epix37:
Nvm, done :). Thank you for the PRs!

andburn:
Sorry, am AFK. The thing that was throwing me with the position was how the unlocking/moving [here](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/Windows/OverlayWindow.xaml.cs#L302) was affected.

Epix37:
Good catch! Fixed that.

PR #1622, opened 2015-10-19
Add an option to disable dialog animation

quark-zju:
Here it is :)

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -775,6 +775,9 @@ namespace Hearthstone_Deck_Tracker
 
 		[DefaultValue(false)]
 		public bool UseAnyUnityWindow = false;
+ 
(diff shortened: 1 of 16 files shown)


Epix37:
Very nice! 

From the [MahApps 1.2 release notes](https://github.com/MahApps/MahApps.Metro/blob/master/docs/release-notes/1.2.0.md) (should be released in a "few weeks")
> * MetroDialogs fixes and changes
  * New property CustomResourceDictionary for DialogSettings

This might allow to easily modify the StoryBoards for the fading animations. I think doing that (maybe speed it up by a factor of 3 or 4) would be a cleaner looking way of doing this than disabling the animations completely.

quark-zju:
So quick! I just made it work for Flyouts as well.

quark-zju:
Yes,  CustomResourceDictionary looks more promising and make animation much shorter works better for most people.

PR #1639, opened 2015-10-22
 Using const string instead of "wild strings", added enum values for rarity, dust value is a property of Card

karimsah:


Epix37:
While I get the idea, I don't quite see the need since the rarity is barely used. So mistakes due to misspelling are fairly unlikely. 

`card.TypeRarity == Card.eTypeRarity.Legendary` is in my opinion also a lot less readable (and more tedious to write) than `card.Rarity == "Legendary"`. 

Why not go with something like
```C#
public class Rarity
{

}

**Comment by Epix37** [comment]
While I get the idea, I don't quite see the need since the rarity is barely used. So mistakes due to misspelling are fairly unlikely. 

`card.TypeRarity == Card.eTypeRarity.Legendary` is in my opinion also a lot less readable (and more tedious to write) than `card.Rarity == "Legendary"`. 

Why not go with something like
```C#
public class Rarity
{
    public static string Legendary { get { return "Legendary" } };
}
```

That would result in a much cleaner `card.Rarity == Rarity.Legendary`

karimsah:
IMO, and according to MS best pratices, strings shouldn't be copy pasted
all around everywhere with the same value.
Since that value is the same everywhere, it should be used in a const.
Using enum values is a lot less tedious to write using Intelisense : you
don't even type it has it's suggested :)

andburn:
To add my two cents. I like the idea of an *enum* for the rarity, but this does seem to be an overly complicated way of doing it.

As a side note, it is generally a good idea to keep the number of commits in a pull request small and clear. You can squash your local commits using *git rebase*, see this [article](https://help.github.com/articles/about-git-rebase/). Its also good practice to stick to the conventions of the repository your contributing to (i.e. the use of tabs or spaces for indentation).

karimsah:
andburn, show me a simple way of doing it :)
Thanks for the squash commits article, i'll try and do that asasp.
As for the tabs and spaces, it was mixed, maybe i should have ignored that
:o

Epix37:
Turns out there is actually a *much* cleaner solution: xml can be deserialized directly to enums. I guess I didn't know that at the time. Doing this with the rest of the card properties would probably be a good idea.

andburn:
@Epix37 strikes again :+1:

karimsah:
Seems good :+1:

Issue #1648, opened 2015-10-24
Frame Level Conflict: Attack Icon and Card Tooltips


Elitesparkle:
We can move the objects forming the overlay, so I decided to move the Attack Icon frames on the right, near my deck list.

After doing this, I've noticed the card tooltips don't appear on top of it, but on bottom.

I suggest you to change the card tooltips frame level to the topmost one, so that it will never be behind something.

Issue #1607, opened 2015-10-13
attack icon does not properly count damage after a minion is polymorphed 


princeboot:
The attack counter keeps the damage total of the minion as if it were not polymorphed. Damage also does not get removed once the polymorph is killed.

Issue #1598, opened 2015-10-12
Attack icon adds weapon damage twice


Epix37:
From the weapon and the hero if weapon is active.

CobraA1:
Yup, I noticed that as well.

![two attack](https://cloud.githubusercontent.com/assets/8573502/10505892/aff3f21a-72d7-11e5-8fb5-57081815a988.png)

PR #1669, opened 2015-10-28
Improve board attack calculation

### LINKED ISSUE #1614: Damage counter counts unspawned tokens from Soul of the Forest
Repro: Play some minions, cast Soul of the Forest. The damage counter for your board shows the combined damage of the minions on the board and the treants that will spawn from them.

Meaning if you had three minions out with 8 combined damage, after playing Soul of the Forest the counter now shows 14.

andburn:
Debug Window:
- fixed the Cards tabs treeview so it saves the expanded collapsed state
- add a damage tab to the debug window, to show the board damage info

Board attack calculation (fixes #1614, #1648, #1607, #1598)
- takes into accout windfury and charge
- only counts minions/weapons able to attack on your turn
- fixes problems with druid "choice" cards and polymorph
- damage icon text goes green when the board kills the other player (doesn't take taunts into account)
- scaled damage icon with height so is smaller at lower resolutions
- change card tooltip zindex so its above icons

Missed the release :disappointed:, stupid windfury weapons caused me so many problems :smile:


Epix37:
The fix seems to be working very well, but all tests in BoardStateTest seem to be failing unless I add `Core.Game = new GameV2();` to the TestInitialize and run them one by one.

andburn:
Damn, must've missed those tests after my last change. Fixed now.

andburn:
Hardcoded a special case for V07TRON, I think it should work ok. Got rid of the green text too.

Epix37:
Looks good :).

PR #1677, opened 2015-10-30
Quick fixes

andburn:
A fix for the board attack when cards are in the wrong zones. Sometimes graveyard, hand and deck. Also, removed the exclusion of Icehowl don't think it was correct in the first place.

Fixed the draw chances 200% thing while I was at it.


Epix37:
:+1:

Issue #1696, opened 2015-11-04
Noble Sacrifice gray out when attacking enemy whit full board


ceceita:
Noble Sacrifice gray out when attacking enemy whit full board , this becouse there is no space in board and secret will not trigger ..

sry for my bad english

PR #1711, opened 2015-11-07
Misc fixes + fix for #1696

ericBG:
Before I start - I'm new to this whole open-source thing, sorry if I messed anything up (and if I did, please tell me - I want to improve!)

Anyways I just decided to make a few small changes on the project to make it slightly more legible while fixing bug #1696 . I've got some awesome ideas coming up (separate secret detection for each secret) but I just wanna know I'm doing this whole open-source thing right first.

Thanks!

### DIFF
--- a/HDTHelper/HDTHelper.csproj
+++ b/HDTHelper/HDTHelper.csproj
@@ -21,6 +21,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <LangVersion>5</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <PlatformTarget>AnyCPU</PlatformTarget>
--- a/HDTTests/HDTTests.csproj
+++ b/HDTTests/HDTTests.csproj
@@ -25,6 +25,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <LangVersion>5</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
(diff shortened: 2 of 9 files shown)


ericBG:
Oh, and also - the reason the lines are - and +'d in commit 22e1ffa is because the line endings are different. Should they be standardised throughout the project?

Epix37:
General feedback: Try to keep your PRs more coherent (? - not sure if that's the word I'm looking for). Two commits that are not at least somewhat related they should probably have their own PR. If there's some change in there that I don't agree with that means more work for at least one of us :).

More detailed:
* Line endings: ~~not sure why the main project is different from the rest.~~ Looks like it's all CR LF, maybe some configuration on your end? You could commit only the one new line (LangVersion) though, that would at least keep it cleaner.
* The color changes: Not a fan of the change. I think it doesn't really accomplish much besides adding more clutter.
* Try to be more descriptive with commit messages than "Fix #1696". With this you need to actually look at the issue/commit to have any idea what was fixed.
  * Does that actually completely fix it? What happens when e.g. Sludge Belcher dies? 
  * If a foreach has more than one line of "content" it should have braces (at least here :)). Coding style is obviously subjective.

ericBG:
So one PR for general cleanup and one for the bugfix, yes?

* Is just LF used in the project? I assumed as it was a windows project we would have CR LF.
* The colour changes were done so that R# can underline the colour (e.g. [here](http://blog.jetbrains.com/dotnet/files/2011/02/css_highlighting.png)). The old format did not allow for that and also made the colours be evaluated at runtime which would be slightly more taxing on the CPU (I know that impact would be negligible)
* #1696:
    * What do you mean when Sludge Belcher dies? As far as I'm aware Noble Sacrifice triggers when there is 6 or less minions and an attack is being made. Those are the only conditions, no?
    * Alright, will fix that.

Epix37:
* It all has CR LF on my end is what I mean.  Are there only LF on git for some reason? I usually push/pull as is. 
* #1696: I'm not sure about the order things are logged in. So I don't know if the following could happen:
  * 1) Sludge dies
  * 2) secrest greyed out
  * 3) 1/2 Slime spawns

Further more I'm not sure if that's maybe even the real behaviour, secret rules are weird...

ericBG:
I did a hex scan a while back, mine are CR LF and yours are LF. 

Hmm I would test myself/look at the logs but my PSU died today :/ Writing this on a mobile.

riQQ:
I checked and it's LF only for me too. Do you have `git config core.autocrlf` set to true?

Epix37:
fixed #1696 with a3f81025cf6ccc36e8c16d46fec092c02ec9063c, merge up to up to 0a45a86.

Project files should now all be CR LF.

Issue #1288, opened 2015-08-21
Suggestion: automatically use no deck in spectator mode


iluvredwall:
Basically as the title states; it would be nice to have an option to automatically enable "use no deck" while spectating games, because it's rare that a friend has the exact same deck that I do.

ericBG:
Is there any way to do this manuall right now?

Epix37:
Manually selecting "no deck"? `MAIN > USE NO DECK` or `"systemtrayicon" > use no deck`

guillermocandia:
I put an option in tracker->general: Use no deck in Spectator mode.

Is it necessary to maintain selected the previous deck?
Could be a new property PreviousActiveDeck in DeckList and handle it in HandleGameEnd()
or just put that info in the config?
https://github.com/guillermocandia/Hearthstone-Deck-Tracker/commit/58ae5e8b3367357ef471f0d5bcf94430bd5f195c

PR #1731, opened 2015-11-11
added the option to automatically use no deck in spectator mode (#1288)

guillermocandia:
Added the option in Tracker->General.
is False by default.
Issue #1288


Epix37:
You should use `Core.MainWindow.SelectDeck(null, true)`, rather than  `DeckList.Instance.ActiveDeck = null;`.

And I think for this to really be useful it needs to re-select the last deck after leaving spectator if there was a deck selected previously. This can be done via `Core.MainWindow.SelectLastUsedDeck()` but HDT needs to keep some state (throughout sessions) of whether a deck was previously active.

guillermocandia:
The deck is re-selected in HandleGameEnd.
The deck is re-selected if the player close HDT during  a game.

Issue #1798, opened 2015-11-28
Crash when editing an arena game 


Sewil:
So I played against an opponent who disconnected and when I looked in the arena games his name wasn't included, so I wanted to edit the game both to add the note that he disconnected and to include his name, but then this happened.

![Image](https://cloud.githubusercontent.com/assets/1498414/11452389/8b55800e-95e6-11e5-8f7d-d49ab9b4a8c4.png)
[Crash report 28112015-1538.txt](https://github.com/Epix37/Hearthstone-Deck-Tracker/files/46361/Crash.report.28112015-1538.txt)

Sewil:
Looked into this some more, and his name was awesomeraff, so no weird characters or anything like that. I think it must be connected to his disconnect, that the servers couldn't receive his name at the time and so it saved a null value or something. I added his name back (Before, it just said <OpponentName />) in the DeckStats.xml file and his class (Shaman) and then I could edit the game again.

Sewil:
I was trying to figure out what caused the crash, and so I removed the class from the game in DeckStats.xml again by setting it to \<OpponentHero />, which again caused the crash, so I'm assuming that's the reason it crashed initially. After his disconnect, it didn't save his class so I couldn't edit the game because it had nothing to put in the class list in the "Add New Game"-window. Maybe a solution would be to add a default "None" value to prevent this?

PR #1800, opened 2015-11-28
Fixed issue #1798

Sewil:
I fixed the issue #1798 where the client would crash when the HeroClass was left as a null value in the AddGameDialog and so when it tried to parse that to the HeroClass enum it just crashed the client. When I played against an opponent who disconnected it never saved the HeroClass value in DeckStats.xml which turned it into a null value. As a solution to this I added a default "None" value for the enum HeroClass and changed the Parse to a TryParse to fetch any null values and turn them into None's.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
+###############################################################################
+* text=auto
+
+###############################################################################
+# Set default behavior for command prompt diff.
+#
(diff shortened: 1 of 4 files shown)


Epix37:
I think adding a none value to the enum is not the right approach here. IIRC there are several things that require a valid hero class (hearthstats syncing being one of them for example). Also changing enum indices always seems a bit dangerous. I don't think they are referred to by index anywhere but better save than sorry (adding it at the end would be saver).

PR #1801, opened 2015-11-29
clear search for export

achansen121:
At least on my computer, decks with singleton cards were not being exported properly because the search box was not being cleared. Simply added an additional clear, and an extra part to the debug line.


Epix37:
Could it be one of these issues? https://github.com/Epix37/Hearthstone-Deck-Tracker/wiki/FAQ#exporting

achansen121:
Yes, that is probably the issue I was having as I have a touchscreen with Windows 8.

I'd assume that this addition wouldn't cause any harm, but well if there's already a solution I guess it's not necessary.

Issue #1794, opened 2015-11-27
Russian letters


Scronheim:
Russian characters are not displayed, win 10, the latest update
http://i.imgur.com/lmjKuGf.png

Floy3D:
Да тажа самая проблема, одни квадраты.

Floy3D:
http://s019.radikal.ru/i605/1511/4e/fac1cd66c2af.png

riQQ:
It's a Windows 10 error (see #1623).

KAMAELUA:
Can you change font in your app, or upload it to app folder and use it instead of windows fonts? That way everyone will have ability to customize their font.

Issue #1783, opened 2015-11-24
Card Names is broken.


grounder78:
i use HDT last version. 
After update to Windows 10 TH2 build 10.0.10586. Card Names in Russian is broken. :(

a build of 10.0.10240 was all right

![hdt](https://cloud.githubusercontent.com/assets/1655120/11365755/ae899a0e-92b9-11e5-95b3-cd1ff798d93e.png)

disshishkov:
:+1:

Epix37:
It's a windows bug (affects more than just HDT). See #1623.

Issue #1756, opened 2015-11-16
Russian translation bugs


HellbringerOnline:
Hi. I have this
![1](https://cloud.githubusercontent.com/assets/14059683/11183567/76203ac4-8c83-11e5-91b6-f0b3218b8761.PNG)
Windows 10.0.10586.3 64bit (12.11.2015)
Tried to change the the file encoding - no results. Others translation still works.

riQQ:
It's a Windows 10 Bug (see #1623).

Issue #1753, opened 2015-11-15
Russian encoding


HEPEBAP:
After new update on windows 10 - 
![hdt_bug](https://cloud.githubusercontent.com/assets/15857634/11169223/813d6218-8bc2-11e5-9438-6deb862980dd.png)

it's problem in Windows, but that problem only in HDT

P.S  That problem only with title
![hdt_bug2](https://cloud.githubusercontent.com/assets/15857634/11169386/0bd8ee5c-8bc7-11e5-84a2-fe1565e8e050.png)

Cu8eR:
It is issue in Win10 according to Google, anyway use another language

riQQ:
See #1623.

grounder78:
I'm using the latest beta HTD 0.13.5.
Name cards can be read, but although font is not as nice as before.
As before, there is a small bug with the symbols " - "  (hyphen)
they appear squares

my previous post #1783

![fontbug](https://cloud.githubusercontent.com/assets/1655120/11753578/eeb9b078-a056-11e5-9053-6bfb589aa17f.PNG)

PR #1805, opened 2015-11-30
Use default font for non-latin languages

andburn:
Tested out problem with the Russian language (#1794, #1783, #1756, #1753, #1623), the OS solutions didn't seem to work for me either.
So, added a new **Font** binding prop on `Card` and `Entity` that only uses *Belwe* if the selected lang  is a *Latin* type language, otherwise it should use the default system font.

There's probably a nicer way, but it works :smile:


Epix37:
Does this issue actually occour with non-latin languages besides russian? That's the only one I've seen this issue with so far.

andburn:
I was able to confirm it for Russian and some Polish characters. The Asian langs worked fine, but since they don't use Belwe anyway I thought maybe group all the non-latin together.

Issue #1762, opened 2015-11-17
Stat Tracking bugs


GiggliG:
I'm seeing a bunch of bugs in the stats tracking, and have been for at least a couple weeks. The stat tracking used to be very reliable, but lately it's become very very unreliable. I'm seeing three bugs, and all of them are frequent enough to be irritating.

The most common bug is that HDT fails to record the opponent name, leaving an empty field. This happens so often that if i play 3 or more games, this will happen at least once.

The next bug is a lot less common, but still frequent enough to be noticed. Now and then when playing ranked, the game is recorded as Casual.

The final bug is the most serious. Every now and then, HDT saves a game as the wrong class. I dont know how often this happens, because its not easy to tell when it has happened. But I watched three games I played this morning, and saw it happen on two of them, in such a way that I can see what is happening.

I played against a paladin, warrior, and hunter, in that order, and looked at the stats after each game. 

After the paladin game, I looked at the stats, and it was recorded correctly (apart from missing the Opponent Name).

After the warrior game, that game was correctly recorded as warrior (no Opponent name, again), but the preceding game against Paladin had changed to warrior class.

So, I manually changed it back to paladin, and then played my third game. That game showed up as hunter, but the 2nd game (warrior) was now displayed as hunter.

So it seems like when HDT is assigning the class, if this bug occurs, it assigns the class to the current game, and the game preceding it.

In all three of these games, the opponent name was blank, so maybe there's a link between those two bugs (and that may be a way of spotting which games have the wrong class recorded).

veri745:
I see all three of these behaviors also

neilrqm:
I see the mode and opponent name bugs as well.  I don't see the bug with class, but I haven't changed class this week and can't remember the ground truth so I probably just haven't noticed.

I also see a few games with the mode set to "None", which don't show up on Hearthstats.  I've attached a screenshot showing my last set of games that demonstrate these bugs, and have verified that the three marked with mode "None" aren't on my game record on Hearthstats.  Also, my Power_old.log file contains the correct value for the most recent instance of the missing name (the others are too old to be in the log file).  I'm not sure how to verify that the log file contains the correct mode.

N.b. that I played a game of Tavern Brawl, probably on the 18th, and played through the new wing of LOE on the 20th.  Other than that I've only played mage in ranked mode.  

[...]

Issue #1653, opened 2015-10-24
Fixing game mode detection


Epix37:
#### Switching to LoadingScreen.log for gamemode detection:
LoadingScreen contains among other the following line, which is logged **before and after** games:
* `LoadingScreen.OnSceneLoaded() - prevMode={MODE} currMode={MODE}`

Where {MODE} is one of the following values:
```
  public enum Mode
  {
    INVALID,
    STARTUP,
    [Description("Login")] LOGIN,
    [Description("Hub")] HUB,
    [Description("Gameplay")] GAMEPLAY,
    [Description("CollectionManager")] COLLECTIONMANAGER,
    [Description("PackOpening")] PACKOPENING,
    [Description("Tournament")] TOURNAMENT, //constructed
    [Description("Friendly")] FRIENDLY,
    [Description("FatalError")] FATAL_ERROR,
    [Description("Draft")] DRAFT, //arena
    [Description("Credits")] CREDITS,
    [Description("Reset")] RESET,
    [Description("Adventure")] ADVENTURE, //practice
    [Description("TavernBrawl")] TAVERN_BRAWL,
  }
```

This should allow for farily easy mode tracking (brawl currently works based on asset unloading) with one exception.

#### Ranked:
Ranked detection seems to be very inconsitent (in some cases?) and even more so is rank detection. Since I can't find anything reliable in the logs, the best solution might be to rely on visual detection, e.g. @andburn's [C# implementaiton of the Track-o-bot method](https://github.com/andburn/hdt-plugin-rankdetection), either with the addition of legend rank detection or in combination with Bob.log's `legend rank X`.

PR #1808, opened 2015-12-01
Image based rank detection

andburn:
My latest attempt to fix rank mode (#1653, #1762). The rank detection method I tried before, turns out to be fairly inconsistent. So, I came up with another one. This time its direct image recognition of the rank "medals", from my testing its 100%. Obviously it'll need further testing with a beta release, but I'm pretty confident about it (assuming blizzard don't do anything stupid). It does require another library [Aforge](http://www.aforgenet.com/), but its not big.

Adds a new class `Utility.RankDetection`, and you call `RankDetection.Match(Bitmap b)` with a screen capture param and returns:
```
RankResult
{
  int Player
  int Opponent
  bool Success
}
```

I didn't want to start dismantling all the existing rank detection stuff (wouldn't know where to start tbh :) ). So, I added the detection into `GameEventHandler.TurnStart` and seems to work.


Epix37:
Amazing! Seems to work very well from what I can tell!

Merged into [rank_detection branch](https://github.com/Epix37/Hearthstone-Deck-Tracker/tree/rank_detection) for now.

Some changes I made:
* `Utility.RankDetection.Range` needs to be a class or implement the base ctor to compile as C#5 (I added the latter)
* Added `Overlay.IsRankCovered()` to make sure you only see the overlay flickering if actually necessary.
* Added a `DetectRanks()` (extracted from TurnStart) call to Game end to deal with instant condeces (that just happened to happen to me while testing :))
* Made `DetectRanks()` async void with a "running" check to not block the main thread with await `RankDetection.Match()`.
* Added friendslist detection to `DetectRanks()`, which waits for the FL to be closed and displays a message in the overlay before running `Match()` (this needs to not block the main thread even more).
* Added a max number of tries to ranked detection if is disrups the user experience (i.e. overlay flickering). We might want to add a general max tries to avoid the overlay flickering x times once a deck list gets too large.
* Completely (I think?) removed the old methods of rank(ed) detection.

andburn:
Nice additions :+1: 
It'll be interesting to see how it works when people test it out

PR #1864, opened 2015-12-13
Stats - win rate with/without coin

Loli-loli-hunter:
Hi, just wondering if you can add a breakdown of win rates starting first / second (with / without coin) to the [Old] Games & Statistics section.


Epix37:
I will be happy to add that with the new constructed stats.

PR #1918, opened 2015-12-23
Add a try-finally block around avenge synchronization code so that we don't get locked out

donaldswartz:
Add a try-finally block around avenge synchronization code so that we
don't get locked out

### DIFF
--- a/Hearthstone Deck Tracker/GameEventHandler.cs
+++ b/Hearthstone Deck Tracker/GameEventHandler.cs
@@ -496,13 +496,19 @@ namespace Hearthstone_Deck_Tracker
 			if(_awaitingAvenge)
 				return;
 			_awaitingAvenge = true;
-			if(_game.OpponentMinionCount == 0)
-				return;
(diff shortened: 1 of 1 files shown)


Epix37:
Indeed nice catch. Any particular reason for try-finally instead of inverting the if?
```C#
if(_game.OpponentMinionCount != 0)
{
	await _game.GameTime.WaitForDuration(AvengeDelay);
	if(_game.OpponentMinionCount - _avengeDeathRattleCount > 0)
		_game.OpponentSecrets.SetZero(CardIds.Secrets.Paladin.Avenge);
}
_awaitingAvenge = false;
_avengeDeathRattleCount = 0;
```

donaldswartz:
In addition to handling the case where opponentMinionCount == 0, this handles the case where there is an exception thrown in WaitForDuration or SetZero which is handled outside. That may not actually be possible right now so if you think inverting the if is more readable, that works, too.

PR #2005, opened 2016-01-20
Fixed infinite loop with whitespaces

massand:


massand:
Please ignore the first two commits from my really old branches. The last commit (https://github.com/massand/Hearthstone-Deck-Tracker/commit/d293b4d5ff6dd194e8e1e57d1bf52d8ed988b4e0) is the only relevant change for the infinite loop fix.

PR #2057, opened 2016-02-01
Input Multiple Files

Goleeb:
This adds the ability to select multiple files in the input file box, and then creates a new deck for each. If more then one file is selected it forces the decks to be auto saved. Check line 241 in the bottom green box.


Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Import.cs
```diff
@@ -198,9 +198,10 @@ public partial class MainWindow
 					if(deck != null)
 					{
 						SetNewDeck(deck);
-						if(Config.Instance.AutoSaveOnImport)
-							SaveDeckWithOverwriteCheck();
-					}
+                        if (Config.Instance.AutoSaveOnImport)
+                            SaveDeckWithOverwriteCheck();
+                                                   
+                    }
```
Epix37: Please change indention to tabs, that removes unnecessary changes like this one.

PR #2062, opened 2016-02-02
Import Multiple Files

Goleeb:
This allows multiple files to be imported.
Forced auto save to allow multiple files at once.
Now using tabs.


Epix37:
:+1:

PR #2068, opened 2016-02-03
add option to assign or change constructed deck URLs

percentcer:
Sometimes you want to update your deck from a different URL than the original import

### DIFF
--- a/Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Edit.cs
+++ b/Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Edit.cs
@@ -317,6 +317,20 @@ namespace Hearthstone_Deck_Tracker.Windows
 			TagControlEdit.SetSelectedTags(deck.Tags);
 		}
 
+		internal async void BtnSetDeckUrl_Click(object sender, RoutedEventArgs e)
+		{
+			var selectedDeck = DeckPickerList.SelectedDecks.FirstOrDefault();
+			if (selectedDeck == null)
+				return;
+
+			var url = await InputDeckURL();
+			if (string.IsNullOrEmpty(url))
+				return;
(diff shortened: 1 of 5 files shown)


Epix37:
Nice idea.

I see two problems right now:

* "rebase" might not be very clear for the average user. Something like "update url" or "set url" is probably easier to understand.
* The menu item is currently missing from the `DECKS` menu in the menu bar.

percentcer:
Ah, good point.  Updated to use the wording "LINK TO NEW URL" and "LINK TO URL" (if the deck does or does not already have a URL, respectively).

Also added the option to the "DECK" menu (required a slight change to the DeckPickerList_OnSelectedDeckChanged call to update the menu items accordingly)

Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Edit.cs
```diff
@@ -317,6 +317,13 @@ internal void BtnEditDeck_Click(object sender, RoutedEventArgs e)
 			TagControlEdit.SetSelectedTags(deck.Tags);
 		}
 
+		internal async void BtnSetDeckUrl_Click(object sender, RoutedEventArgs e)
+		{
+			var selectedDeck = DeckPickerList.SelectedDecks.FirstOrDefault();
+			selectedDeck.Url = await InputDeckURL();
```
Epix37: One more thing: selectedDeck *could* in theory be null here. I think in reality there is no way for this method to be called without a deck being selected, but `FirstOrDefault()` should be follow by a null check nonetheless.

percentcer:
Added a deck null check and emptystring check to the BtnSetDeckUrl_Click function

Epix37:
Perfect! :+1:

PR #2071, opened 2016-02-05
Export Save As XML now supports multiple selected Decks

Goleeb:
Saves decks by name when multiple decks are selected.
Might want an option to turn this function on or off.

### DIFF
--- a/Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Export.cs
+++ b/Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Export.cs
@@ -135,12 +135,33 @@ namespace Hearthstone_Deck_Tracker.Windows
 			var deck = DeckPickerList.SelectedDecks.FirstOrDefault();
 			if(deck == null)
 				return;
-			var fileName = Helper.ShowSaveFileDialog(Helper.RemoveInvalidFileNameChars(deck.Name), "xml");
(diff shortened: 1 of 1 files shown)


Hearthstone Deck Tracker/Windows/MainWindow/MainWindow_Export.cs
```diff
@@ -135,12 +135,33 @@ private void BtnExport_Click(object sender, RoutedEventArgs e)
 			var deck = DeckPickerList.SelectedDecks.FirstOrDefault();
 			if(deck == null)
 				return;
-			var fileName = Helper.ShowSaveFileDialog(Helper.RemoveInvalidFileNameChars(deck.Name), "xml");
-			if(fileName == null)
-				return;
-			XmlManager<Deck>.Save(fileName, deck.GetSelectedDeckVersion());
-			await this.ShowSavedFileMessage(fileName);
-			Logger.WriteLine("Saved " + deck.GetSelectedDeckVersion().GetDeckInfo() + " to file: " + fileName, "Export");
+			System.Windows.Forms.SaveFileDialog dialog = new System.Windows.Forms.SaveFileDialog();
```
Goleeb: Adding using System.Windows.Forms caused two Clipboard calls to be ambiguous. So I accessed the SaveFileDialog like this.

PR #2073, opened 2016-02-06
Add flavor text on MouseOver to overlay

Epix37:


Tenchi2xh:
This is a really nice feature! I love it! Flavor text used to be so inaccessible and now you can read it while you play, should be part of the core game :D

Issue #1387, opened 2015-09-03
Suggestion: show/hide news bar


Elitesparkle:
My news bar is currently hidden but I'd like to see it again.

After looking in the options, I can't find any check box to activate it.

Can you add it?
Thanks.

Epix37:
Sure.

For now you can turn it on again by setting ` <IgnoreNewsId>` in the config to `-1`.

Elitesparkle:
> For now you can turn it on again by setting <IgnoreNewsId> in the config to -1.

When I've seen there were the value 7, I didn't know which value to set to enable it again.

Thanks.

culinko:
The option has been added into Tracker -> Settings -> Show news bar.

PR #2079, opened 2016-02-09
add option for showing or hiding news bar

JDurman:
Issue #1387

### DIFF
--- a/Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
+++ b/Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
@@ -45,7 +45,7 @@ namespace Hearthstone_Deck_Tracker.FlyoutControls.Options.Tracker
 			CheckBoxShowLoginDialog.IsChecked = Config.Instance.ShowLoginDialog;
 			CheckBoxShowSplashScreen.IsChecked = Config.Instance.ShowSplashScreen;
 			CheckboxStartWithWindows.IsChecked = Config.Instance.StartWithWindows;
-			CheckBoxAnalytics.IsChecked = Config.Instance.GoogleAnalytics;
+			CheckBoxAnalytics.IsChecked = Config.Instance.GoogleAnalytics;			
 
 			if(!Helper.EventKeys.Contains(Config.Instance.KeyPressOnGameStart))
 				Config.Instance.KeyPressOnGameStart = "None";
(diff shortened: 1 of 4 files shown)


Hearthstone Deck Tracker/Config.cs
```diff
@@ -396,7 +396,10 @@ public class Config
 		[DefaultValue(-1)]
 		public int IgnoreNewsId = -1;
 
-		[DefaultValue(true)]
+        [DefaultValue(true)]
+        public bool ShowNewsBar = true;
```
Epix37: There is no need for a new config value, checking `IgnoreNewsId != -1` is sufficient.

Hearthstone Deck Tracker/Core.cs
```diff
@@ -118,8 +118,10 @@ public static void Initialize()
 			MainWindow.Options.OptionsTrackerPlugins.Load();
 			PluginManager.Instance.StartUpdateAsync();
 
-			UpdateOverlayAsync();
-			NewsUpdater.UpdateAsync();
+            if (Config.Instance.ShowNewsBar)
```
Epix37: Checking this removed the ability for me to remotely enable the news bar again, which might be useful is some cases.

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
```diff
@@ -357,5 +359,33 @@ private void CheckBoxAnalytics_OnUnchecked(object sender, RoutedEventArgs e)
...
+            if (!_initialized)
+                return;
+            Config.Instance.IgnoreNewsId = -1;
+            Config.Instance.ShowNewsBar = true;
+            Config.Save();
+            Utility.NewsUpdater.UpdateAsync();
+        }
+
+        private void CheckboxShowNewsBar_Unchecked(object sender, RoutedEventArgs e)
+        {
```
Epix37: Extracting `MainWindow.BtnCloseNews_OnClick()` to a new method and calling it here would be much cleaner.

Epix37:
Two more things:
- [Please use tabs instead of spaces.](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/CONTRIBUTING.md)
- Currently, closing the news bar via its integrated close button would not update the checkbox in options.

JDurman:
Thanks for the feedback.  Yeah when I was doing it I wasn't sure if it should be a permanent disable independent of the ignore. I can definitely see why you want to remotely enable it so I'll just implement the IgnoreId functionality.  Sorry about the spaces my visual Studio was converting my tabs to spaces.

JDurman:
Alright It's working now and I have rebased it onto the most recent commit.

Epix37:
Merged with https://github.com/Epix37/Hearthstone-Deck-Tracker/commit/4733755a3a74cfac53843e40b1d9d68e58243a24

PR #2159, opened 2016-02-29
Make the formatting show up for flavour texts.

ericBG:
Using the `Inlines` property of the TextBlock, I made it so that the flavour text of the cards shows with the formatting properties of the flavour text - either bold or in italics.


Epix37:
Now for the slightly more nitpicky things:
- there are still a few unnecessary formatting changes.
- the `Hearthstone Deck Tracker.csproj` change should probably be in the first commit
- the first commit message should describe what the class does on it's own (especially since it has it's own commit), something along the lines of "adds functionality to parse formatted strings on TextBlock"

ericBG:
Alright, on the rebase right now. Believe it or not I've never used Vim before so I have no clue how to use it.

* Yeah I tried just doing tab to spaces using the VS "Tabify setting", I'll try to revert them manually... It doesn't seem to like tabifying XAML

* Whoops, it was meant to be.

* Should that be in the extended commit message?

ericBG:
Remade this with another repo, git was messing with my brain.

PR #2211, opened 2016-03-14
Card bar themes

andburn:
First phase of the card bar themes. No actual themes yet, mostly restructuring, moving the images into folders only including whats necessary in the project. Also, added the latest card images and updated the frame graphics and coordinates:
preview
Just want to make sure this doesn't cause any problems before I push anymore.


Epix37:
Looking very nice! Doesn't cause any problems as far as I can see. 

Could you optipng the images?

andburn:
Crap, I actually did optimize them but it looks like I must have messed up the rebase/reordering and the originals got pushed. Will fix it.

andburn:
There's only really two things that are different to the previous frames:
- added a white created icon (see sunwalker above)
- and changed the count text to turn to a green nine when the count goes over nine, instead of the legendary star

over_nine

Epix37:
- Like the created icon a lot. 
- Agree with moving away from the legendary star for >9 but not sure about making it a green 9. The green doesn't really have any intuitive meaning and "9" is just wrong :). Maybe a `*` instead?

andburn:
Yeah, I agree - I couldn't think of anything better at the time. Maybe a small `*` or `+` at the top right of the 9 could look good. I'll try something out now.

Epix37:
Do you have an eta? I would like to include this in 0.13.16 if possible.

Epix37:
Just noticed this: not sure how I feel about the different colors for legend-star and numbers. I *do* like the orange though.

andburn:
How about this for >9 
over_nine

I can change the numbers to same color as star if you want

Epix37:
Looks good! Not sure if the star or the number color should change, but I think they should be the same either way.

andburn:
So just had to check but they are actually different in the game: 

ingame

hdt

I'll still change them though if you want :smile:

Epix37:
Huh. Leave it then I suppose :smile:.

andburn:
The more I looked at the odder it looked so made them all yellow. Updated. :laughing: 

yellow

Epix37:
Merged and [fixed an issue where legendaries would not display the card count](https://github.com/Epix37/Hearthstone-Deck-Tracker/commit/793caea0ef3a92f5881a28640f3dedf91335452f#diff-ae0d8885db8ec8ed448eac3254a493b4R120).

andburn:
Damn it, I always miss something. Thanks.

PR #2241, opened 2016-03-19
Add card set icons to deck edit mode

andburn:
Replace the text based set indicators with [icons I made](https://github.com/andburn/hs-icons).

set_icons


Epix37:
very nice!

Issue #2243, opened 2016-03-19
can not detected china region


chucklu:
this method UpdateOverlayAsync under Hearthstone-Deck-Tracker\Hearthstone Deck Tracker\Core.cs
Game.CurrentRegion = Helper.GetCurrentRegion();

you can not get the china region from the config file
I found the config file under the C:\Users\Administrator\AppData\Roaming\Battle.net
It will not be updated when i switch to china region,the region config is just valid for US,EU,KR

However,we can get the correct region by check the ServerAddress
 Game.CurrentRegion = Helper.GetRegionByServerIp(Game.MetaData.ServerAddress);

chucklu:
I will try to fix this

chucklu:
finally,i found we must detect the region by server addres before we save the game record
detect the region only for the game start is not enough

Epix37:
Fixed with v0.14.0.

PR #2244, opened 2016-03-19
detect the  correct region 

chucklu:
First, detect the region by server address 

If we can not detect the region by server address, then detect the region by config file

If we can not detect the region by config file,then log the server address.
if we can detect the region by config file ,then log the server address and the region detected by the config file


chucklu:
This request is trying to fix this issue https://github.com/Epix37/Hearthstone-Deck-Tracker/issues/2243

Epix37:
* I don't want to rely on using IPs if not necessary, so for anything but china config should be the preferred method.
* You don't need to specifically call `Helper.GetCurrentRegion` at this point, it is called on game start and stored in `Game.CurrentRegion`. The region can't change while the game is running.
* Why the changes to `Helper.GetRegionByServerIp`?

chucklu:
Helper.GetCurrentRegion is called when we can not detect the region by ip

The changes make to  Helper.GetRegionByServerIp is good for debug,when i try to debug the method,
I want to set the breakpoint at the last line of the method.

PR #2308, opened 2016-04-03
String Comparison Clean-up

Jecoms:
Cleaned up string comparisons to use the safer .Equals or
string.IsNullOrWhitespace().

### DIFF
--- a/HDTTests/HearthStats/ApiTest.cs
+++ b/HDTTests/HearthStats/ApiTest.cs
@@ -40,7 +40,7 @@ namespace HDTTests.HearthStats
 			_deck1.Note = "test note 1";
 			_deck1.Tags = new List<string> {"testtag1", "testtag2"};
 			_deck1.Class = "Mage";
-			foreach(var card in Database.GetActualCards().Where(c => c.PlayerClass == "Mage").Take(30))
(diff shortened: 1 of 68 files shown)


Jecoms:
Other suggestions I have would be to implement more enums when referencing titles from a set. i.e. The different card sets are referenced by their string name instead of an enum.

Epix37:
`string` is a special case when it comes to the  `==` operator, if both sides are of type `string`, it always resolves to `Equals`. 

Besides that, using `str.Equals(str2)` is vulnerable to throw NullReferenceExceptions. The "safe" way to do it would be to use `string.Equals(str1, str2)`, but that is *exactly* what the `==` operator does with two strings.

Epix37:
You can find the HearthDb library here: https://github.com/HearthSim/HearthDb

[CardSet](https://github.com/HearthSim/HearthDb/blob/master/HearthDb/Enums/CardSet.cs) has no names that are of any help, but HDT has `HearthDbConverter.SetConverter` method that turns those into the proper strings.

Jecoms:
Ah. That was my Java bleeding through. I suppose I should have realized it wouldn't be an issue since you don't use the String object class. Another reason to like C# more!

I'll look into contributing more meaningful code after I've spent more time in it.

PR #2311, opened 2016-04-04
Adjust board damage text for ChunkFive

andburn:
IMO, the board damage text looks too big for the graphic with ChunkFive, so I adjust the values to make it a bit smaller.
attack

*Unrelated, but added the wild and standard icons they'll be of use sometime maybe*


Epix37:
:+1:

PR #2313, opened 2016-04-05
Added CardSet Enum to replace strings

Jecoms:
Also added Enum extension method to convert to natural English with spaces.


jleclanche:
There's already an enum in [HearthDB](https://github.com/HearthSim/HearthDb/blob/master/HearthDb/Enums/CardSet.cs) which should probably be used instead.

Jecoms:
I realize this, but I didn't want to couple it to his reference library. Since a conversion dictionary was used, I figured it was intended to keep them separate. Either way, an enum is an improvement.

Epix37:
I agree with jleclanche here, enum duplication is not the way to go. At least if it's completely decoupled from HearthDb. Some wrapper that makes the set names more readable could be an option.

Enums should also not have names containing underscores and words starting with lowercase. Solving this with attributes seems like the cleanest way to me: [see here for reference](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/Enums/DisplayedStats.cs) - [converter](https://github.com/Epix37/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/Enums/EnumDescriptionConverter.cs).

The point of HearthDb as a separate library was to be able to make it reusable and license it under MIT. Using it is the whole point of having it :).

Jecoms:
I can refactor to use the HearthDB in that case. I think there was one area that required a comparison to a more readable string so I could handle that with the description attribute as you suggested. I'll make the changes this afternoon.

Jecoms:
At this point we can't use the HearthDB CardSet enum because it doesn't have a constant for Whispers of the Old Gods. Further, are the enums derived from the game files? They are very poorly named (GvG is PE1?). It would be much better and more usable if they were named properly.

If the plan is to reduce duplication of enums between HearthDB and HDT, there is currently duplication of the Rarity enum in both. The HDT enum could be replaced by the HearthDB enum.

Epix37:
Yes, HearthDb uses the same enums Hearthstone does.

PR #2320, opened 2016-04-07
Use HearthDB.Enums.Rarity

Jecoms:
HDT-Rarity Annotated with Obsolete.

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone/Card.cs
+++ b/Hearthstone Deck Tracker/Hearthstone/Card.cs
@@ -13,7 +13,6 @@ using Hearthstone_Deck_Tracker.Annotations;
 using Hearthstone_Deck_Tracker.Utility;
 using Hearthstone_Deck_Tracker.Utility.Logging;
 using Hearthstone_Deck_Tracker.Utility.Themes;
-using Rarity = Hearthstone_Deck_Tracker.Enums.Rarity;
 
 #endregion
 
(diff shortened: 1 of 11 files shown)


Hearthstone Deck Tracker/Hearthstone/Card.cs
```diff
@@ -115,7 +114,7 @@ public Card(HearthDb.Card dbCard)
 			Id = dbCard.Id;
 			Count = 1;
 			PlayerClass = HearthDbConverter.ConvertClass(dbCard.Class);
-			Rarity = HearthDbConverter.RariryConverter(dbCard.Rarity);
+             Rarity = dbCard.Rarity;
```
Jecoms: I don't know what is wrong with the white space upon submitting a pull request. It looks fine in VS.

riQQ: Check `Edit > Advanced > View White Space`. It's a tab vs space issue.

Epix37:
I can squish the commits for you, no problem.

I'm not sure making the old enum obsolete actually solves the problem though, I was waiting on a response from @ko-vasilev. :)

ko-vasilev:
@Epix37 Sure it breaks the plug-in, but it's not hard to modify it to use another type for rarities.

Epix37:
I was hoping there would be a way we could avoid breaking it, but I guess not huh?

ko-vasilev:
I mean, if the Rarity property of Card class stays of the same type and this enum will still be usable it should not break the plug-in. Otherwise I don't think there is a way to keep plug-in compatible.

Epix37:
Hm yeah... I can't really think of a way to make it work other than introducing a new property for the HearthDb Rarity and leaving the old one in as obsolete - which is not a good soluition.

This is not a change I want to break plugin compatibility for I think.

My suggestions now would be to leave this open until I have to break compatibility anyway (if I decide to update MahApps at some point for example).

ko-vasilev:
I mean, it's not a big deal to change the plugin code to use another type for Rarity.

Epix37:
I'm mainly concerned about the user experience. Everyone would have to manually update again.

ericBG:
Maybe it's time to try and make an automatic plugins manager? In fact, that seems like an idea...

Epix37:
might as well merge this now I guess

Epix37:
Merged with https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/294d737e6482ec897f3d8ba81d3859cff20dfbce.

Issue #2036, opened 2016-01-29
Enhance-o-Mechano - Not showing Windfury effect text in replay


jmac-slash0:
When playing Enhance-o-Mechano the Divine Shield and Taunt effects are displayed as text on the cards in replays. However, no text is present for cards given the Windfury effect. You can see the issue in the attached replay on turn 9.

[Occ(Warlock) vs Twfic(Warlock) 2120-280116.zip](https://github.com/Epix37/Hearthstone-Deck-Tracker/files/110285/Occ.Warlock.vs.Twfic.Warlock.2120-280116.zip)

azeier:
Local replay support has now been fully removed.

PR #2462, opened 2016-05-02
Fix for issue #2036 - Enhance-o-Mechano - Not showing Windfury effect text in replay

craigt00:
Hi.  I've fixed issue #2036, which was caused because 'Windfury' was missing as a tag within the Entity.Effects() method.  I've also cleaned up a few hardcoded empty strings and newlines.

Thanks.

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone/Entities/Entity.cs
+++ b/Hearthstone Deck Tracker/Hearthstone/Entities/Entity.cs
@@ -168,21 +168,21 @@ namespace Hearthstone_Deck_Tracker.Hearthstone.Entities
 		{
 			get
 			{
-				var effects = "";
+				var effects = string.Empty;
 				if(HasTag(GameTag.DIVINE_SHIELD))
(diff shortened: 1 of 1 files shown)


Epix37:
Could you squash that last commit into the previous two?

craigt00:
No problem.  Just to be clear, you just want all three commits squashed into one?

Epix37:
Nono, just the last one. The first two as separate commits is perfect.

craigt00:
Hopefully it should be ready now.  Thanks!

Epix37:
Close enough :). I moved the indention fix for windfury into the first comit.

Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/710b12c20b423fcd31fbc3c54aca47f5255ef372 and https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/791ee907279225e97d79fc2cd0cc26ed614341af.

:+1:

Issue #2396, opened 2016-04-27
Request： Add Deck Title , Deck Tag, Dust cost ,etc. to Deck Screenshot


everchump:
![deck](http://i65.tinypic.com/2v9w1g6.jpg)

The pic is modified from the topdecks (URL [HERE](http://www.hearthstonetopdecks.com/deck-builder/#?deck_id=21913)).

May you add an option to add these stuffs disappearing in the screened deck pic ?

- Deck Title：Deck Name & Hero background pic. It's better if  you can select the hero bg pics.    
- Deck Tag(Date/Note）：You can add a tag or date here ，for example，“Midrange".  
- Dust Cost：Dust cost of the deck，as a new player, it's useful.

Elitesparkle:
Nice idea, but keep it optional, please.

andburn:
This is pretty cool idea.
@Elitesparkle why would you want it optional out of interest, specifically when would the extra info interfere with the way you use the screenshot?

Cu8eR:
Really good idea! and useful!

Elitesparkle:
I have some ideas about things to add:
- Format: based on cards in it, show if the deck is Standard legal or not;
- Sets: as you do in the tracker, show an icon for each cards set needed to build it;
- Rank: show the best Rank you have reached with that deck version;
- Wins/Losses (Percentage): show how many games you have won and lost, win rate included.

> @Elitesparkle why would you want it optional out of interest, specifically when would the extra info interfere with the way you use the screenshot?

Sometimes you don't need to show too many information (e.g. you don't know the deck author, you didn't give a name to your new deck yet, you don't want to show your Rank and/or win rate).

PR #2475, opened 2016-05-05
Improve deck screenshot

andburn:
Opening this for discussion, comments etc. as requested in #2396 added some extra info to the deck screenshot.

wild std

**Requested**
- [x] Title with hero background
- [x] Deck tag
- [x] Dust cost
- [x] Format standard or wild (icon and text)
- [x] Sets used
- ~~Rank (*Not sure this belongs on a deck screenshot*)~~
- ~~Wins/Losses (Percentage) (*Like rank doesn't seem right*)~~
- [x] Make all extra info optional (*More options?*)

Uses a new window instead of using the `PlayerWindow`. Also, added a bit of branding at the end a good addition I think.

**Issues**
- ~~Gets the first tag (or class if no tags), *first tag could be anything, favorite etc. not really much can do about that*~~
- ~~To keep the set icons from changing color with the Mahapps theme, copied them to the xaml. *Could do it a better way, maybe*~~
- ~~Its another chunk of code that needs to be updated with new sets~~


Epix37:
Amazing, as always!

Two things don't quite fit in I think: 
- the pattern seems too noisy compared to the rest
- the black background of the bottom most panel. Using the same dark gray as the fade/countbox might be better.


> Rank (Not sure this belongs on a deck screenshot)
 Wins/Losses (Percentage) (Like rank doesn't seem right)

I agree, doesn't belong.

> Make all extra info optional (More options?)

All or nothing is fine. Either adding a second screenshot button, something like `SCREENSHOT (WITH INFO)`, or a dialog with both options and remember functionality should work.

> Gets the first tag (or class if no tags), first tag could be anything, favorite etc. not really much can do about that

There's only a very limited number of "highlevel' deck types (Aggro, Control, Midrange, ...) only checking for those should work. Maybe including Brawl too.

> Its another chunk of code that needs to be updated with new sets 

Adding a `DeckSets` control sounds like the best solution here. Something you can just pass the deck.

everchump:
I think the set icons should be adding on the card bg pic behind the card name.

Epix37:
You mean on each card? I disagree. That would adds a lot of noise to the screenshot, make it inconsistent with the rest of the app and also make it a lot harder to identify the included sets (assuming they are removed from the bottom panel).

everchump:
You can set an option to add or not add the set icon.

As a new player , may be you don't know which set a card  come from.

If the set icon is putting on each card ,it's useful.

It doesn't matter how much sets the deck contain.

Like the Hearthhead's deck theme,[WOTOG Tempo Mage](http://www.hearthhead.com/deck=165108/wotog-tempo-mage), for example, the set icon is putting before the card name.

Epix37:
My point still stands and I still disagree. I want to have as little options as possible on this, it's not a feature that needs customization.

culinko:
@andburn I love it! Amazing work! I agree with everything Epix said. Rank and W/L don't belong to the screenshot. Checking for only specific 'high-level' deck tags (Aggro/Face, Midrange, Control, Combo, Mill, Fatigue, Ramp, Tempo) and stuff like Brawl (or even Arena) would make the most sense, because I wouldn't want the decks having "Favorite" or "Missing Cards" tags exported :P

Cu8eR:
@culinko - I agree with you - just tags should be fine - constructed/arena/brawl/adventure?

another seoarated tags should be aggro/control/tempo

Maybe what about deck version? Show it under the name

Epix37:
Arena, Brawl and Adventure (if tagged as such?) should replace the Standard/Wild text. Icons for each would be nice.

I don't think the deck version adds any value here. These would mostly be created to be shared with other people, who don't care about your local deck revisions.

andburn:
**Updated**

- changed the patterns as suggested, does looks better

decks

- icons for arena and brawl

spiders

- created set icons control
- use only predefined tags (*Midrange, Aggro, Control, Tempo, Combo*)
- add option to export menu "*Screenshot (with info)*"

Cu8eR:
Just one thing, what about PvE (Adventure) icon? Maybe number of spells/minions? Just suggestion from hearthpwn

I like your latest version, really good! -thank you

Epix37:
Adventure icon might be nice, not sure how many of those decks are actually tagged as such - might have too check for multiple tags (adventure, pve, maybe others). 

I think numbers of spells/minions would be too much here. 


Other than that, @andburn, should I look into merging this or are there any outstanding changes?

andburn:
Added an adventure icon for completeness (uses *adventure* or *pve* tags).

adv_icon

Should be OK to megre now.

Epix37:
I think we might want to check for any tag containing "pve"  or "adventure". IIRC, the default hearthpwn tag is "pve adventure". 

I'm gonna go through the code tomorrow - maybe wait for that so you don't have to push new commits 10 times :).

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Windows.Media.Imaging;
+using Hearthstone_Deck_Tracker.Hearthstone;
+using Hearthstone_Deck_Tracker.Utility.Extensions;
+using static System.Windows.Visibility;
+
+namespace Hearthstone_Deck_Tracker.Windows
+{
+	public partial class DeckScreenshotWindow
+	{
+		private bool _showDeckOnly;
```
Epix37: Can be a local variable.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		public double DeckMaxHeight => ActualHeight - LabelsHeight;
+
+		public double LabelsHeight => DeckTitleContainer.ActualHeight + DeckFormatPanel.ActualHeight
+			+ SetDustPanel.ActualHeight + BrandContainer.ActualHeight + 42;
+
+		public void UpdateWindowHeight()
+		{
+			this.Height = StackPanelMain.ActualHeight + 42;
```
Epix37: No need for `this`. Should be converted to an expression (`public void UpdateWindowHeight() => Height = StackPanelMain.ActualHeight + 42;`)

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		private ImageBrush DeckHeaderBackground(string deckClass)
+		{
+			var heroId = ClassToID(deckClass);
+			DrawingGroup DrawingGroup = new DrawingGroup();
```
Epix37: Should use var and name should start with lowercase.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+			return new ImageBrush {
+				ImageSource = new DrawingImage(DrawingGroup),
+				AlignmentX = AlignmentX.Left,
+				Stretch = Stretch.UniformToFill
+			};
+		}
+
+		private string GetTagText(Deck deck)
+		{
+			List<string> predefined = new List<string>() {
```
Epix37: Should use var.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+				RectIconAdventure.Visibility = Visible;
+			else if(deck.StandardViable)
+				RectIconStandard.Visibility = Visible;
+			else
+				RectIconWild.Visibility = Visible;
+		}
+
+		private int TotalDust(Deck deck)
+		{
+			List<string> nonCraftableSets = new List<string>() {
```
Epix37: Should use var.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+				"Curse of Naxxramas",
+				"Blackrock Mountain",
+				"League of Explorers",
+				"Basic"
+			};
+			List<string> nonCraftableCards = new List<string>() {
```
Epix37: Should use var, use `HearthDb.CardIds` instead of raw card id strings.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+					return "HERO_07";
+
+				case "warrior":
+				default:
+					return "HERO_01";
+			}
+		}
+
+		internal void Shutdown()
+		{
```
Epix37: Should be converted to an expression.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+
+			return deck.Cards
+				.Where(c => !nonCraftableSets.Contains(c.Set) && !nonCraftableCards.Contains(c.Id))
+				.Sum(c => c.DustCost * c.Count);
+		}
+
+		private string ClassToID(string klass)
+		{
+			switch(klass.ToLowerInvariant())
+			{
```
Epix37: Should use `HearthDb.CardIds` instead of raw card id strings. (E.g. `Collectible.Mage.JainaProudmoore`)

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+			RectIconArena.Visibility = Collapsed;
+			RectIconBrawl.Visibility = Collapsed;
+			RectIconAdventure.Visibility = Collapsed;
+
+			var allTags = deck.TagList.ToLowerInvariant();
+			if(deck.IsArenaDeck)
+				RectIconArena.Visibility = Visible;
+			else if(allTags.Contains("brawl"))
+				RectIconBrawl.Visibility = Visible;
+			else if(allTags.Contains("adventure") || allTags.Contains("pve"))
```
Epix37: Needs to check if any tag contains the phrase "adventure" or "pve".

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+			RectIconStandard.Visibility = Collapsed;
+			RectIconWild.Visibility = Collapsed;
+			RectIconArena.Visibility = Collapsed;
+			RectIconBrawl.Visibility = Collapsed;
+			RectIconAdventure.Visibility = Collapsed;
+
+			var allTags = deck.TagList.ToLowerInvariant();
+			if(deck.IsArenaDeck)
+				RectIconArena.Visibility = Visible;
+			else if(allTags.Contains("brawl"))
```
Epix37: Needs to check if any tag contains the phrase "brawl". "Taverl brawl" would fail here.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+			var allTags = deck.TagList.ToLowerInvariant();
+			if(deck.IsArenaDeck)
+				return "Arena";
+			else if(allTags.Contains("brawl"))
+				return "Brawl";
+			else if(allTags.Contains("adventure") || allTags.Contains("pve"))
+				return "Adventure";
+			else if(deck.StandardViable)
+				return "Standard";
+			else
```
Epix37: Redundant else keywords.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,187 @@
...
+				"Control",
+				"Tempo",
+				"Combo"
+			};
+			if(deck.Tags.Count > 0)
+			{
+				foreach(var tag in predefined)
+				{
+					var found = deck.Tags.Find(
+						f => f.ToLowerInvariant().Replace("-", "") == tag.ToLowerInvariant());
```
Epix37: Checking for contains rather than equality might also be better here.

Hearthstone Deck Tracker/Windows/MainWindow.NewDeck.cs
```diff
@@ -306,12 +306,7 @@ private void AddCardToDeck(Card card)
 
 		private void UpdateExpansionIcons()
 		{
-			RectIconOg.Visibility = _newDeck?.ContainsSet("Whispers of the Old Gods") ?? false ? Visible : Collapsed;
-			RectIconLoe.Visibility = _newDeck?.ContainsSet("League of Explorers") ?? false ? Visible : Collapsed;
-			RectIconTgt.Visibility = _newDeck?.ContainsSet("The Grand Tournament") ?? false ? Visible : Collapsed;
-			RectIconBrm.Visibility = _newDeck?.ContainsSet("Blackrock Mountain") ?? false ? Visible : Collapsed;
-			RectIconGvg.Visibility = _newDeck?.ContainsSet("Goblins vs Gnomes") ?? false ? Visible : Collapsed;
-			RectIconNaxx.Visibility = _newDeck?.ContainsSet("Curse of Naxxramas") ?? false ? Visible : Collapsed;
+			SetIcons.Update(_newDeck);
```
Epix37: Should be converted to an expression.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
andburn: Its using [Deck.TagList](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/Hearthstone/Deck.cs#L315), so it will catch *brawl* in any tag. Same for *adventure*, *pve* below.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
andburn: Looks like this is leftover from me copying `PlayerWindow`, will get rid of it and just call `Close()` directly.

andburn:
Updated as separate commit so diff is clear. Will squash it when your happy.

everchump:
So does  the extra info support local languages ?

"Aggro"  ---> "速攻“ ， for example.

It seems as  it didn't support local languages.

Hearthstone Deck Tracker/Windows/DeckScreenshotWindow.xaml.cs
```diff
@@ -0,0 +1,172 @@
...
+						return tag;
+
+			return deck.Class;
+		}
+
+		private string GetFormatText(Deck deck)
+		{
+			if(deck.IsArenaDeck)
+				return "Arena";
+			else if(_allTags.Contains("brawl"))
```
Epix37: Sorry, I meant all of the 'else' here are unnecessary, not just the last one :).

Epix37:
I don't think supporting extra language is feasible here. There's a good chance the meta types will change with time - I don't really feel like maintaining that in 16 languages. Possibly when general localization is added.

@andburn Add `using static HearthDb.CardIds.Collectible` in `DeckScreenshotWindow` and I think we're gtm!

everchump:
I mean is there a way to set the extra info manually like the Deck names or Deck Tags in other languages by the way you input it yourself？

Epix37:
Not for now,

andburn:
Ok should be good now :shipit:

Epix37:
:+1:
Merged with https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/729fbb6ac951a6ad6d681b949b24993f5e4dd49c.

everchump:
- So 14.8 delete the Light themes ? 
- How to change hero bg pics? Is there no way to change it ?
- `Make all extra info optional (More options?)` I know you don't want add more stuffs into option. But if i can set which extra info or Info bar to disappear will be cool ...

PR #2556, opened 2016-05-31
Fix deck screenshot distortion at low resolutions

andburn:
So, I was thinking about this and realized there's no need to actually use a `Window` because we aren't using `PlayerWindow` anymore. Changing it from a Window to a `UserControl` and using `RenderTargetBitmap` it can render in memory and isn't limited by the users screen size. It does use the manual height calculation again, but it works this time :wink:


Hearthstone Deck Tracker/Windows/MainWindow.Export.cs
```diff
@@ -75,19 +77,23 @@ private void BtnExport_Click(object sender, RoutedEventArgs e)
...
-			var pngEncoder = Helper.ScreenshotDeck(screenShotWindow.StackPanelMain, 96, 96, deck.Name);
-			screenShotWindow.Close();
-			await SaveOrUploadScreenshot(pngEncoder, deck.Name);
+			var deck = selectedDeck.GetSelectedDeckVersion();
+			var cards = 35 * deck.Cards.Count;
+			var height = (deckOnly ? 0 : 124) + cards;
+
+			DeckView control = new DeckView(deck, deckOnly);
+			control.Measure(new System.Windows.Size(219, height));
+			control.Arrange(new Rect(new System.Windows.Size(219, height)));
```
Epix37: `System.Windows` qualifier is unnecessary for this and the line above.

Epix37:
Very nice! Just fix that one minor thing, rebase on master again are we're gtm!

andburn:
Done

Issue #909, opened 2015-04-22
Add options to auto-archive or delete decks tagged as "arena" after 3 losses and/or 12 wins


NightKev:
Separate options for each possibility would be nice.

mkrain:
I like this, currently I'm doing this manually.

jimrrchen:
+1 on this. This can just be an option that user can toggle on/off

Epix37:
Implemented with #2575.

PR #2575, opened 2016-06-04
Add option to auto-archive arena decks after the run is over

aalear:
This partially addresses enhancement request #909: an option to automatically archive arena decks when the run is over is added to the Tracker settings under "General".

Automatic deletion would be easy to tack on as well, but I'm not sure that's actually a good idea in practice since deletion, unlike archiving, is a destructive action with various options possible (e.g. delete stats vs keep stats, etc). Popping up the usual warning/confirmation screens would mitigate the data loss risk, but then it's no longer really automatic anyway.

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -94,6 +94,9 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue(67.5)]
 		public double AttackIconOpponentHorizontalPosition = 67.5;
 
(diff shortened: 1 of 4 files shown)


NightKev:
IMO you can consider #909 closed if this is merged, I've realized that it's better to just archive all arena decks.

Epix37:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/f027ed681c1d321cbaf6e3d7999494e9845b7a40.

Closes #909.

PR #2613, opened 2016-06-17
Added option to select Hearthstone Log directory

becelot:
Currently, the Hearthstone Log directory is only dependent on the Hearthstone directory (logs can be found in \Logs). 

The purpose of this change is to provide functionality for Hearthstone multiboxing having a separate tracker window for every hearthstone instance. This is a feature I'm currently working on which will be provided as a plugin at some point.

Changes:
- Logs directory can now be changed to a different Log directory.

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -427,6 +427,9 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue(@"C:\Program Files (x86)\Hearthstone")]
 		public string HearthstoneDirectory = @"C:\Program Files (x86)\Hearthstone";
 
+		[DefaultValue("Logs")]
+		public string HearthstoneLogsDirectory = "Logs";
+
 		[DefaultValue("Hearthstone")]
 		public string HearthstoneWindowName = "Hearthstone";
 
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -428,7 +428,7 @@ namespace Hearthstone_Deck_Tracker
 		public string HearthstoneDirectory = @"C:\Program Files (x86)\Hearthstone";
 
 		[DefaultValue("Logs")]
-		public string HearthstoneLogsDirectory = "Logs";
+		public string HearthstoneLogsDirectoryName = "Logs";
 
 		[DefaultValue("Hearthstone")]
 		public string HearthstoneWindowName = "Hearthstone";
--- a/Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
+++ b/Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
@@ -415,5 +415,32 @@ namespace Hearthstone_Deck_Tracker.FlyoutControls.Options.Tracker
 			Config.Instance.AlternativeScreenCapture = false;
 			Config.Save();
 		}
+
+		private async void ButtonHearthstoneLogsDirectory_Click(object sender, RoutedEventArgs e)
+		{
+			var dialog = new FolderBrowserDialog();
+			dialog.SelectedPath = Config.Instance.HearthstoneDirectory;
+			var dialogResult = dialog.ShowDialog();
+
+			if (dialogResult == DialogResult.OK)
+			{
+				//Logs directory needs to be a child directory in Hearthstone directory
+				if (!dialog.SelectedPath.StartsWith(Config.Instance.HearthstoneDirectory + @"\"))
+				{
+					await Core.MainWindow.ShowMessage("Invalid argument", "Selected directory not in Hearthstone directory!");
+					return;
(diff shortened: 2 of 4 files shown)


Epix37:
- How exactly does multiboxing work with Hearthstone? From this it looks like you can use the same installation but have it output the logs to a different folder?

- Without knowing what this option is for, it's not clear it's is supposed to do.

- Even with multiple HDT installation you can currently only run it once (I think) - requiring multiple installation is not a good solution either way though. For this to work properly we probably want some kind of "user profile system" that allows you to have multiple HDT configurations. 

- The overlay is currently attached to the first Hearthstone instance found. This needs a system to mark a Hearthstone proc (window handle) as "in use". Ideally it can, at the same time, get the HS installation from the proc (assuming multiboxing requires multiple installations), so that starting a second HDT instance would automatically select the 2nd window and the correspondig log without any configuration.

becelot:
- Only one installation is required for multiboxing. But using symlinks, we are able to output the logs to different directories (setting the Logs folder as a directory symlink to another directory). The symlink can be changed to a different directory every time another instance of hearthstone is stared.
- Currently, I'm using multiple HDT installations with a script to sync all installations afterwards and automatically create symlinks and start Hearthstone for different regions.
- HDT can be run as often as you want. It shuts down if it cannot get write access to its own log file. However, changing the data directory for HDT creates separate log files for all installations. You can have as many instances of HDT running at the same time as you want.
- A "user profile system" is currently just different HDT tracker installations that work independently by setting the data directory locally and not use a global configuration from AppData.
- Note that while HDT indeed attaches to the first hearthstone instance found, this is always the instance that is "the most recently started". Thus by starting hearthstone and the corresponding HDT alternating, every HDT instance attaches to the correct Hearthstone instance.

I agree that this is not an optimal solution. However, it works (I'm running that setup for multiple month now). Allowing multiboxing with a single instance of HDT requires some major reworking of existing systems, which is not feasible for the amount of users that will actively use that feature. I just proposed this change because some people over at reddit asked about HDT multiboxing and I wanted to provide my workaround that does the job for multiple month now (and because I'm lazy and I try to update HDT as infrequently as possible to not have to reimplement this feature).

Ultimately, it is not meant to be a major feature, but to provide a workaround for users that want to multibox, since the implementation of a real solution is just too much work.

Epix37:
Fair enough, that sounds like it works a lot better than I expected. Two very minor things and it'll be gtm then.

Could you add instructions to the wiki on how to set it up?

Also: the email you used to commit is no associated with your account. If you want the commits to be associated with your account you need to add that email in settings.

Hearthstone Deck Tracker/Config.cs
```diff
@@ -151,7 +151,7 @@ public class Config
 		[DefaultValue(true)]
 		public bool ConstructedAutoImportNew = true;
 
-		[DefaultValue(true)]
+		[DefaultValue(true)]
```
Epix37: There should not be a change here.

Probably inconsistent line endings :/? I need to reformat everything at some point.

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
```diff
@@ -415,5 +415,34 @@ private void CheckboxAlternativeScreenCapture_Unchecked(object sender, RoutedEve
...
+				//Logs directory needs to be a child directory in Hearthstone directory
+				if (!dialog.SelectedPath.StartsWith(Config.Instance.HearthstoneDirectory + @"\"))
+				{
+					await Core.MainWindow.ShowMessage("Invalid argument", "Selected directory not in Hearthstone directory!");
+					return;
+				}
+
+				//Check if same path selected (no restart required)
+				if (Config.Instance.HearthstoneLogsDirectoryName.Equals(dialog.SelectedPath))
+				{
```
Epix37: No need for braces around a single line.

Hearthstone Deck Tracker/Config.cs
becelot: yea, inconsistent line ending. That specific line uses LF line ending, while all other lines end with CRLF. Should I revert this change or leave it (for consistency?).

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerSettings.xaml.cs
becelot: Fixed

becelot:
Everything should be fixed, reverted that line ending change. Also connected my github account to my other email address, accidentally pushed using the wrong SSH key.

Epix37:
Could you fixup the last 2 commits into `98989dd` and `a62e12b` and rebase everything on master? Should be good to go then.

becelot:
Okay, should be done. Didn't know about that fixup option provided by git :)

Epix37:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/compare/26f2f5d370ed...f1ad26762648.

PR #2658, opened 2016-07-08
Fix potential error importing from clipboard

andburn:
A very small thing I noticed. 
If you import a deck from the clipboard and a card has a comma in its name and you use a count with it (e.g. 'Ragnaros, Lightlord x1'), it won't be recognized.
Since only legendaries have commas currently, this will probably never actually cause a problem though :smile:.


Epix37:
60 lines of tests for one minor change, very nice! :D

PR #2709, opened 2016-07-27
ImportAll checkbox fix

pedrotst:
The importAll checkbox was working oddly because was not nicely binded
with the other checkbox.

### DIFF
--- a/Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
+++ b/Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
@@ -48,7 +48,7 @@ namespace Hearthstone_Deck_Tracker.Controls
 			get { return _text; }
 			set
 			{
-				if(_text == value)
+				if (_text == value)
 					return;
 				_text = value;
 				OnPropertyChanged();
@@ -60,7 +60,7 @@ namespace Hearthstone_Deck_Tracker.Controls
 			get { return _text; }
 			set
 			{
-				if(_text == value)
+				if (_text == value)
 					return;
 				_text = value;
 				OnPropertyChanged();
@@ -85,7 +85,7 @@ namespace Hearthstone_Deck_Tracker.Controls
 		public void SetDecks(List<ImportedDeck> decks)
 		{
 			Decks.Clear();
-			foreach(var deck in decks)
+			foreach (var deck in decks)
 				Decks.Add(deck);
 			_ready = decks.Any();
 			Text = NoDecksFoundText;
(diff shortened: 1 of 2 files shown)


pedrotst:
CheckBoxImport_OnChecked could have probably be done by adding some kind of counter binded to the CheckBoxImport. This would lead to extra perfomance.
Maybe I'll try that later.

Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
```diff
@@ -112,6 +112,31 @@ private void CheckBoxImportAll_OnModified(object sender, RoutedEventArgs e)
 			}
 		}
 
+		private void CheckBoxImport_OnChecked(object sender, RoutedEventArgs e)
+		{
+			int modifiedCount = 0;
```
Epix37: Should be `var modifiedCount = 0;` or `int modifiedCount;`.

Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
```diff
@@ -112,6 +112,31 @@ private void CheckBoxImportAll_OnModified(object sender, RoutedEventArgs e)
...
+
+			if (ItemsControl == null)
+				return;
+
+			foreach(var item in ItemsControl.Items)
+			{
+				var c = (ContentPresenter)ItemsControl.ItemContainerGenerator.ContainerFromItem(item);
+				var chk = (CheckBox)c.ContentTemplate.FindName("CheckBoxImport", c);
+                if (chk.IsChecked.Value == true)
```
Epix37: ` == true` is redundant when looking at a boolean.
This should be either `chk.IsChecked.HasValue && chk.IsChecked.Value`, `chk.IsChecked ?? false` or `chk.IsChecked == true` (in this case it's a `bool?`).

Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
```diff
@@ -112,6 +112,31 @@ private void CheckBoxImportAll_OnModified(object sender, RoutedEventArgs e)
...
+					modifiedCount++;
+			}
+
+            if (modifiedCount == ItemsControl.Items.Count)
+                CheckBoxImportAll.IsChecked = true;
+
+		}
+
+        private void CheckBoxImport_OnUnchecked(object sender, RoutedEventArgs e)
+        {
```
Epix37: This can be converted to an expression. [See here](https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6#expression-bodies-on-method-like-members).

Hearthstone Deck Tracker/Controls/DeckImportingControl.xaml.cs
```diff
@@ -112,6 +112,31 @@ private void CheckBoxImportAll_OnModified(object sender, RoutedEventArgs e)
...
+		{
+			int modifiedCount = 0;
+
+			if (ItemsControl == null)
+				return;
+
+			foreach(var item in ItemsControl.Items)
+			{
+				var c = (ContentPresenter)ItemsControl.ItemContainerGenerator.ContainerFromItem(item);
+				var chk = (CheckBox)c.ContentTemplate.FindName("CheckBoxImport", c);
```
Epix37: Since you are using this in OnModified as well, `c` and `chk` can be extracted to a `GetCheckBox` method, making both methods a bit cleaner.

Epix37:
There are also some spaces instead of tabs for whitespace again ;).

Let me know if there are any other changes you want to make to it before I merge. I'll probably squash this into the earlier commit.

pedrotst:
Can you link me a convention guideline so I stop annoying you with so many simple things? :)

Epix37:
The general stuff is in here: https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/CONTRIBUTING.md

Other than that there's no guideline at the moment :/.

pedrotst:
the .xaml is all with spaces, should I retab it?

Epix37:
Good point, no, that just applies for .cs files.

pedrotst:
I think this feature is good enough now, I don't think I'll be changing it again, go ahead and squash it.

I guess I'll get the used with the rules as we go along and I learn more c#  ^^

Epix37:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/defc66d893f8c4229cc4df642e49b7a76fba94d1.

Did some last cleanup. `ItemsControl` should never be null so I removed those checks.

PR #2725, opened 2016-08-02
Hsimport newsite marduck

LuizSD:
Insert a new site for import system.

### DIFF
--- a/Hearthstone Deck Tracker/Importing/Websites/Marduktv.cs
+++ b/Hearthstone Deck Tracker/Importing/Websites/Marduktv.cs
@@ -1,0 +1,51 @@
+﻿#region
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Web;
+using Hearthstone_Deck_Tracker.Hearthstone;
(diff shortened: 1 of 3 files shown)


Hearthstone Deck Tracker/Importing/Websites/Marduktv.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+	public static class Marduktv
+	{
+		public static async Task<Deck> Import(string url)
+		{
+			return await ImportMarduktv(url);
+		}
+
+		private static async Task<Deck> ImportMarduktv(string url)
+		{
```
azeier: There is no need to have this as a separate method, it can be directly in `Import`.

azeier:
I think once `ImportMarduktv` in inlined this should be good. 

After that, please [fixup](https://help.github.com/articles/about-git-rebase/) your commits into one and rebase it on master.

Once all that is done, you can just force push to this branch and the PR will automatically update.

azeier:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/88f375759dbd71250052e553ccf0c053fc99862f

PR #2735, opened 2016-08-06
Load plugins with embedded dependencies

andburn:
I've been looking for a way to merge plugin dependencies into a single dll. Tried a few different ways with no success. Finally got [Costura.Fody](https://github.com/Fody/Costura) to work with libraries, but I need to add a little [hack](https://github.com/Fody/Costura/issues/104) to the plugin manager. The merged assemblies will throw an exception on `Assembly.GetTypes()`, this can be solved by first trying to instantiate a known dummy class added to the default namespace of the plugin, literally
```csharp
public class MergedTrigger { } // could be called anything, just picked this
```
For assemblies without such a class it will have no effect, doesn't have any other side effects either.


azeier:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/50f0a9af49edcece0c6251c704af2aa0c5dea078

PR #2758, opened 2016-08-16
Add Tribal filter to deck creation menu

ddipaolo:
Simply adds another filter for deck creation to filter by the Tribal tags.  The HearthDb stuff calls this "Race" but everywhere else I've seen Tribe.  Changing it to Race would be trivial obviously, but with the addition of more sets and more emphasis on some synergies I think it is a good option to have.  For example, if you want to make a Beast Druid with the new cards you can filter for cards of Beast type now.  Or Pirates, or Murlocs, or Dragons, etc.


azeier:
This functionality already exists by enabling `options (advanced) > tracker > general > full text search in deck builder`, which allows to search for anything in the description text / race or type.

Jonysegal:
Closing due it being two years old

PR #2789, opened 2016-09-01
Fix card images sporadically not rendering

### LINKED ISSUE #757: Images showing improperly in deck view
While editing or making a deck, or even just viewing, if you scroll, often the cards will look like in the image unless you remove/add something in the deck. Of which it goes back to normal for a short time. This happens with almost all of my decks after the update.

capture

I don't know what causes this, but it's ridiculously annoying.

riQQ:
Fixes #757


azeier:
Do you know why this fixes it or why it happens in the first place?

riQQ:
Not really, I'd love to know, too.

azeier:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/598aa1b6fba73d470d70f08bfe823efb8f4555fb

PR #2792, opened 2016-09-04
add chinese localization

jindaxia:


Hearthstone Deck Tracker/Enums/Language.cs
```diff
@@ -7,6 +7,8 @@ public enum Language
 		[Description("English")]
 		enUS,
 		[Description("German")]
-		deDE
+		deDE,
+		[Description("���ģ����壩")]
```
jleclanche: Encoding on the file doesn't look correct.

sdbds:
[translation by bdsqlsz.zip](https://github.com/HearthSim/Hearthstone-Deck-Tracker/files/453893/translation.by.bdsqlsz.zip)

PR #2793, opened 2016-09-04
update for Chinese Localization

sdbds:
translation some string to Chinese

### DIFF
--- a/Hearthstone Deck Tracker/Properties/Strings.zh-CN.resx
+++ b/Hearthstone Deck Tracker/Properties/Strings.zh-CN.resx
@@ -397,13 +397,13 @@
     <value>筛选</value>
   </data>
   <data name="MainWindow_DeckBuilder_Filter_Cost" xml:space="preserve">
-    <value>消耗</value>
+    <value>费用</value>
   </data>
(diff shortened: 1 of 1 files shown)


Hearthstone Deck Tracker/Properties/Strings.zh-CN.resx
```diff
@@ -118,6 +118,1120 @@
...
+    <value>另存为</value>
+  </data>
+  <data name="MainWindow_DeckBuilder_Button_Cancel" xml:space="preserve">
+    <value>取消</value>
+  </data>
+  <data name="MainWindow_DeckBuilder_Button_Filter" xml:space="preserve">
+    <value>筛选</value>
+  </data>
+  <data name="MainWindow_DeckBuilder_Filter_Cost" xml:space="preserve">
+    <value>消耗</value>
```
jindaxia: I think 费用 would be better

jindaxia: I think 费用 would be better

azeier:
I moved the Strings files to a separate repo: https://github.com/HearthSim/HDT-Localization

Could you create a new PR over there?

sdbds:
OK，got it.

PR #2798, opened 2016-09-05
[BoradDamage] exclude PLAYER entity ensure we can get the correct minions count

jindaxia:


azeier:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/aa6255648904090aca18407c926cc12101e67587

PR #2813, opened 2016-09-10
some localization fix

jindaxia:


Hearthstone Deck Tracker/Utility/Helper.cs
```diff
@@ -54,8 +54,8 @@ public static class Helper
 		public static readonly Dictionary<string, string> LanguageDict = new Dictionary<string, string>
 		{
 			{"English", "enUS"},
-			{"Chinese (China)", "zhCN"},
-			{"Chinese (Taiwan)", "zhTW"},
+			{"简体中文", "zhCN"},
+			{"繁體中文", "zhTW"},
```
jleclanche: I don't think this dict is supposed to be translated.

jindaxia: but many Chinese friends of mine think it is needed
and the options of language in BattleNet client are also showed as many language
options

jleclanche: The dropdown should be translated yes, but I don't know if this particular dict should be. @azeier ?

Hearthstone Deck Tracker/Controls/ElementSorterItem.xaml.cs
```diff
@@ -21,7 +21,33 @@ public partial class ElementSorterItem
...
+				case "Wins":
+					CheckBox.Content = Utility.LocUtil.Get("Options_Overlay_SortedElement_Wins");
+					break;
+				case "Win Rate":
+					CheckBox.Content = Utility.LocUtil.Get("Options_Overlay_SortedElement_WinRate");
+					break;
+				default:
+					CheckBox.Content = name;
+					break;
+			}
```
azeier: Did you test this? I think switching languages might cause problems with the way this feature is implemented.

Hearthstone Deck Tracker/Utility/Helper.cs
azeier: There should be a better way of doing it (e.g. using `CultureInfo.GetCultureInfo(...).NativeName`), but I think translating this might be fine for now.

Hearthstone Deck Tracker/Controls/ElementSorterItem.xaml.cs
jindaxia: @azeier , you are right , must restart to see the changes

azeier: That makes sense, I expected it more to basically reset when changing the language (or possibly crash)  because it uses different strings.

jindaxia: try to use <lex:EnumComboBox> instead of combobox

azeier:
Merged first 3 commits, implemented the ordering in a more reliable way: https://github.com/HearthSim/Hearthstone-Deck-Tracker/compare/d8fdf4c862b1...4a76c01b3e92

PR #2815, opened 2016-09-11
Importing Support

clemson81:
Please add support for importing from:

http://www.vicioussyndicate.com/


azeier:
I think you made the pull request from the wrong branch.

That aside, we would much prefer websites implementing this: https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/2769 (It's supported by the current hsreplay build)

Issue #2753, opened 2016-08-13
Secret Tracker: Competitive Spirit mistakenly greyed out


jedimaster421:
When an opponent starts his turn with a Paladin secret active, and it doesn't trigger, the Secret Tracker will grey out Competitive Spirit even when the opponent has no minions on board. However, Competitive Spirit will only trigger if there are minions it can buff, so it shouldn't be greyed out in that case.

azeier:
I am not able to reproduce this, seems to work fine.

manuel-delverme:
happens to me aswell

manuel-delverme:
[log_2382016-193347.txt](https://github.com/HearthSim/Hearthstone-Deck-Tracker/files/433158/log_2382016-193347.txt)
happened again

riQQ:
Can you upload the corresponding .hdtreplay from the Replays sub-folder?

manuel-delverme:
i sent @azeier the power.log is that enough?

UrzaMTG:
I also see this issue regularly. I've attached a ZIP with the replay file.
[UrzaMTG(Shaman) vs Witles(Paladin) 1921-230816.zip](https://github.com/HearthSim/Hearthstone-Deck-Tracker/files/433809/UrzaMTG.Shaman.vs.Witles.Paladin.1921-230816.zip)

riQQ:
@manuel-delverme that's fine.

rrwoods:
Apologies for submitting a duplicate.  As this issue isn't open, I didn't see it in my search.

azeier:
reopening this so we stop getting duplicates...

vladertel:
Apologies for submitting a duplicate.

PR #2818, opened 2016-09-13
Prevent Competitive Spirit from being prematurely greyed out

aalear:
This addresses issue #2753 where if the opponent played Competitive Spirit, it would be greyed out regardless of opponent minion presence.


azeier:
Should already be fixed with https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/80aa285d4cb6c3c3fbb95a2e866a47c96086dda3

aalear:
Fair enough. I thought I tested it without the minion count check, but it's possible I got my prod and local versions confused. Closing this PR, sorry for the noise.

PR #2827, opened 2016-09-19
fix color scheme option

jindaxia:


azeier:
I moved the Strings files to a separate repo: https://github.com/HearthSim/HDT-Localization

Could you create a new PR over there?

PR #2831, opened 2016-09-21
fix active/use

jindaxia:


azeier:
Merged as https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/c3e5d3e9172052ff23550205c59c294ef9d29eee with slightly different commit message.

PR #2842, opened 2016-09-24
Hsreplay

panos754:
Could u add a feature in replay mode where u can make a Note/Reminder of that game and player. Then the note will appear on your next encounter with the same opponent.
For example, i start a new game and match up to a cntrl warrior named KLOL and he has Deathwing. I have made a note " uses Deathwing and baron" from our previous encounters.  When the game is starting after mulligan that small note will appear reminding me that i have played against him 3 times  and the msg " uses Deathwing and baron"

ty in advance


jleclanche:
Do you mean on HDT or on the website?

I'm closing this because you mistakenly filed a pull request instead of an issue.

PR #2887, opened 2016-10-12
Pull latest changes from official branch

irufus:


irufus:
whoops. wrong request

Issue #2888, opened 2016-10-12
HDT displays flavor text tooltip even on cards without any flavor text


jleclanche:
Such as The Coin.

PR #2922, opened 2016-10-29
Don't show flavor text on minion without it

judge2020:
addressing issue #2888

### DIFF
--- a/Hearthstone Deck Tracker/Windows/OverlayWindow.Tooltips.cs
+++ b/Hearthstone Deck Tracker/Windows/OverlayWindow.Tooltips.cs
@@ -313,7 +313,7 @@ namespace Hearthstone_Deck_Tracker.Windows
 		{
 			try
 			{
-				if(!Config.Instance.ShowFlavorText || entity?.Card == null)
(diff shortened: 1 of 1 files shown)


judge2020:
preview:

https://www.youtube.com/watch?v=DIl_p8a3Obg

Hearthstone Deck Tracker/Windows/OverlayWindow.Tooltips.cs
```diff
@@ -313,7 +313,7 @@ private void SetFlavorTextEntity(Entity entity)
 		{
 			try
 			{
-				if(!Config.Instance.ShowFlavorText || entity?.Card == null)
+				if(!Config.Instance.ShowFlavorText || entity?.Card?.FormattedFlavorText == "")
```
azeier: You need to use `string.IsNullOrEmpty` here. If entity or card are null this will return false.

judge2020: I originally thought the `?` after `card` would return false if card was null, however I agree `IsNullOrEmpty` is probably much cleaner.

PR #2934, opened 2016-11-08
Add section about bootstrapping

mikler:
New contributors might not notice the existence of bootstrap.bat and that
can lead to longer onboarding times. The fix is to mention this file in
contributing section.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,6 +5,10 @@
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


azeier:
I think this belongs in the readme rather than in contributing.

mikler:
@azeier The way the things are right now README.md looks like more of an info file for consumers, rather than for potential contributors.

PR #2954, opened 2016-11-18
Focus search box when form loads

judge2020:
When search is clicked, the text box will auto-focus so the user doesn't have to click on it themselves. 

As always, demonstration: https://www.youtube.com/watch?v=2x4lAIbAoQo

### DIFF
--- a/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
+++ b/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
@@ -20,7 +20,6 @@ namespace Hearthstone_Deck_Tracker.FlyoutControls.Options
 		public OptionsSearch()
 		{
 			InitializeComponent();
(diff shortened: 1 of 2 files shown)


Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -20,6 +20,12 @@ public partial class OptionsSearch : UserControl
 		public OptionsSearch()
 		{
 			InitializeComponent();
+			TextBoxSearch.Loaded += TextBoxSearchLoaded;
```
azeier: Please use xaml to subscribe to default events.

judge2020:
Loaded seems to be called each time that form gains focus after being on another page, clicking away from the form unloads is while clicking back calls `Loaded` again.

So each time you click search from another page the textbox will focus.

azeier:
Alright, looks good then.

PR #2969, opened 2016-11-26
Search tab updates with search results as user types

judge2020:
The search tab will automatically update with search results as user types, as long as they have three or more characters in the text box. 

The search button can still be used for reasons of searching with 2 or less characters, however will probably not be used.

### DIFF
--- a/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
+++ b/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
@@ -1,5 +1,6 @@
 ﻿#region
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows;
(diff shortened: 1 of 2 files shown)


Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -103,5 +104,12 @@ public UserControlWrapper(UserControl userControl, string name)
 			public UserControl UserControl { get; set; }
 			public string Name { get; set; }
 		}
-	}
+
+        private void TextBoxSearch_TextChanged(object sender, TextChangedEventArgs e)
+        {
+            if (TextBoxSearch.Text.Length < 3)
+                return;
+            ButtonSearch_OnClick(ButtonSearch, e);
```
azeier: You can just call `UpdateSearchResult(TextBoxSearch.Text)` here, no need to go through the event method.

PR #2979, opened 2016-11-30
Cleanup events from disabled addons.

RedHatter:
Events registered by disabled plugins should be removed from the ActionList. Disabled plugins shouldn't be able to receive events; if the plugin is disabled it shouldn't be doing anything. In addition the ActionList contains a reference back to the method the plugin passed in which, as there is no Remove method, will prevent the Object where the method is defined from being disposed when the addon is disabled creating a likely memory leak.


azeier:
:+1: Thanks!

PR #2996, opened 2016-12-03
export deck

judge2020:
What's needed: localization update for these strings (which probably won't even work since the hearthpwn database is English) 

More deck provider support would be welcome.

Also, for some reason VS decided to do some spaces formatting.


judge2020:
Going to instead make this a plugin

PR #3005, opened 2016-12-04
Reword file-cleanup message

isloat:
"do just do" and "talk too long" stuck out to me when first installing HDT today.


azeier:
You got that message when first installing HDT? That should not happen.

Did you have an older version installed at somepoint?

isloat:
Now that I think of it, I did install an older version a couple years ago.

PR #3027, opened 2016-12-13
Reload Plugins

judge2020:
Adds a reload plugins button under "plugins folder".

Currently it restarts HDT, this is likely a temporary solution until plugins are loaded within their own AppDomain.

It is only visible with "advanced options" turned on, as this feature is going to be most useful for developers testing plugins. A user likely won't need to continuously reload plugins.

Edit: strings will need to be updated. The name could either be "reload plugins" or "restart HDT", whichever is preferred.


azeier:
Hmm...

If this is mainly intended for plugin dev purposes, wouldn't writing a plugin that hooks into `MainWindow.Closing` and starts another instance be more convenient?

PR #3051, opened 2016-12-27
when cancelled export screenshot,  send the screenshot to clipboard

jindaxia:
it's usefull to copy the image to some IMs


azeier:
This should not happen on cancel.

jindaxia:
ok, where can i put the button on?

azeier:
Functionality has been added with the new dialog.

PR #3057, opened 2016-12-31
Fix Mana Crystals Import Issue

Gamic:
When Mana Crystals does not show a set of cards in either the class card or the neutral card sections of the website, a null pointer exception was thrown. 

This change should fix that issue.


azeier:
Thanks :+1:

Issue #3056, opened 2016-12-31
Unit Test "EliteHearthstone" Failing


Gamic:
The EliteHearthstone Unit Test is attempting to pull a deck list from http://www.elitehearthstone.net/deck-7918-patron-warrior/. However, when I run the test no deck list is pulled back from the site. When I try to follow the link I'm taken to a parked hosting page. (The same seems to be true for Google Speed Test on that page). 

Is this just an issue with me, or is the elite hearthstone website no longer available?

riQQ:
It's unavailable for me, too.

jleclanche:
Yeah it's not available. I think we can remove support for it.

PR #3059, opened 2016-12-31
Remove support for EliteHearthstone

judge2020:
http://www.elitehearthstone.net/ no longer is a Hearthstone website. This removes the test and support for EliteHearthstone.


judge2020:
#3056

PR #3093, opened 2017-01-28
translation into japanese

milkshake-HS:
According to https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/453c99331ea704ffabb95abc4549e1270ae2e406,
I fix some files for translation into japanese.
Is that all needed for translation?


judge2020:
Hey, instead of making a PR here wait for your [PR in the HDT-localization](https://github.com/HearthSim/HDT-Localization/pull/27) to be approved and the HDT maintainer @azeier will add in the functionality for it. 

The build failed at first because it clones the contents of HDT-localization, however, it's not in that repository yet.

milkshake-HS:
Ok, I understand it.

PR #3109, opened 2017-02-15
Added support for Ukrainian (uk) language

olloff:
As for https://github.com/HearthSim/HDT-Localization/pull/28


riQQ:
Maybe using the Mono resgen works: https://linux.die.net/man/1/resgen.

olloff:
fe2e84a needed as a workaround for next issue: In "Options > Overlay > General" field "Hide Card-marks" set to 114px. 200+ needed to add a reasonable description, esp. with localized strings.

Vladsg:
maybe ua language? uk - united kingdom

olloff:
@Vladsg this is language tag, it can be either uk or uk-UA, see https://msdn.microsoft.com/en-us/library/cc233982.aspx

azeier:
Oh that's interesting actually. Could you change it to uk-UA (here and on the localization repo)? That would require a bit less custom implementations to make two codes work.

olloff:
@azeier Hope I did it right.
Sorry about the mess with intendations.

azeier:
Thank you! :)

Merged (without merge commits): https://github.com/HearthSim/Hearthstone-Deck-Tracker/compare/v1.1.7.0...b8baf0af4775982a3bdd41106dfb0a8df80513fd.


Also made the options menu a bit wider and added auto sizing for the options tree to help with clipping.

PR #3111, opened 2017-02-15
Fix certain plugins when dragging-and-dropping

judge2020:
sorry for the commit spam,

Before, with drag-and-drop, a directory would be for the plugin, which would prevent certain plugins like ArenaTracker from working. This includes subdirectories of subdirectories in the plugin search as a fix.


judge2020:
This is a potentially better method at doing the extracting with drag-and-drop.

when reading a zip file, the name property of an entry seems to be null when it's a folder. I'm not sure why and if this is a bug, but as of now, it seems to work. This will also make sure the second entry contains a slash, otherwise the root contains multiple files,

judge2020:
Thanks on the discord!

This now will extract to the plugins folder regardless and then check if the directory is a single folder. if so, it will delete the folder which it just extracted to and re-extract correctly.

This fixes the problem with Arena Tracker (and various other plugins) while still maintaining the directory structure.

Issue #2806, opened 2016-09-07
Export to imgur fails


spacian:
When I try to export a screenshot to imgur, HDT says "uploading" and then I get the message that the tmp file is "saved to ~filepath~" in the Appdata folder. I don't get the link to imgur anymore. Full Reinstall (uninstalling on Windows via programs and features, deleting system files) didn't solve the issue. Deactivating the Firewall didn't solve the issue either. Running out of ideas from my side, would appreciate any tip though :)

HDT version 1.0.5.

azeier:
`12:38:36 AM|Error|Imgur.Upload >> System.Net.WebException: The remote server returned an error: (429).`

Looks like imgur is getting too many requests on the API key HDT is using. The only reliable way to fix this would be adding your own API key in the config file (`%AppData%/HearthstoneDeckTracker/config.xml`), which you can get from https://api.imgur.com.

pquerner:
If you know what http status code is returned when an API key is overused, cant you point that out to the user? Or is it just for the techies?

judge2020:
We could do a string#Contains('429') and show an error if the key has hit the rate limit. @azeier

PR #3115, opened 2017-02-25
Plugin Browser

judge2020:
This PR introduces a new feature: the plugin browser and plugin updating.

There is a new interface for plugins to use, `IUpdatable`, which will allow them to provide the URL to their GitHub repo where it will check for updates to the plugin. This interface is not required, as to not break old plugins, but can be added to improve functionality.

There are two new buttons on the main installed plugins page, "bugs/issues" and "details". details will bring you to the readme, and bugs/issues will bring you to the plugin's issues page. The buttons are disabled if no repo URL is provided for the plugin or `IUpdatable` is not present.

The "available" plugin list downloads json from https://github.com/HearthSim/HDT-Plugins/blob/master/plugins.json and parses it into the list. 

One possible problem is GitHub will rate limit the user if they have too many plugins installed and constantly check for new plugin updates. The rate limit is 60 requests and resets within a few hours. Only one request is made with the "available" plugins list.

This probably can be helped by hiding the JSON file behind a server, like hsdecktracker.net.

The description of the plugin in "updates" will likely be markdown on Github but will be plain text in the box. Not a big issue.

Localization for the "bugs/issues" and "details" buttons is certainly possible, but cross-language plugins may be scarce enough for this to be a priority.

Also, if we want to keep the plugins folder button, I'm not sure where a good place to put it would be.

---

Notify me if there are any questions and I am certainly open to changes.


judge2020:
Here is a video of the features so far - uninstalling plugins, installing new plugins, Implementing IUpdatable in a plugin, and updating a plugin. 

https://youtu.be/gPtcY1cbaqc

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+
+		private List<Plugin> GetUpdates()
+		{
+			try
+			{
+				Log.Info("Checking for updates to installed plugins.");
+				var updateList = new List<Plugin>();
+				if(ListBoxPlugins.Items.IsEmpty)
+					return null;
+				var wc = new WebClient();
```
azeier: Object initializer, `using`

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+					{
+						Log.Info($"{plugin.Name} cannot be checked for updates. Invalid repo url.");
+						continue;
+					}
+					var release = releaseUrl(plugin.Repourl);
+					var releaseData = wc.DownloadString(release);
+					var latestVersion = Version.Parse(GetVersion(releaseData));
+					if(plugin.Plugin.Version >= latestVersion)
+						continue;
+					var Plugin = new Plugin
```
azeier: Variable should be lowercase

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+		private string GetBody(string releaseData) => JObject.Parse(releaseData)["body"].ToString();
+
+
+		private bool rateLimitHit()
+		{
+			var wc = new WebClient();
+			wc.Headers["User-Agent"] = $"Hearthstone Deck Tracker {Core.Version} @ Hearthsim";
+			var json = JObject.Parse(wc.DownloadString("https://api.github.com/rate_limit"));
+			wc.Dispose();
+			if(json["rate"]["remaining"].ToString() == "0")
```
azeier: You can just return this condition

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+			}
+		}
+		private string GetAuthor(string releaseData) => JObject.Parse(releaseData)["author"]["login"].ToString();
+
+		private string GetVersion(string releaseData) => JObject.Parse(releaseData)["tag_name"].ToString().Replace("v", "");
+
+		private string GetBody(string releaseData) => JObject.Parse(releaseData)["body"].ToString();
+
+
+		private bool rateLimitHit()
```
azeier: Methods should be CamelCase

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+		{
+			var wc = new WebClient();
```
azeier: `using`, object initializer

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+			wc.Headers["User-Agent"] = $"Hearthstone Deck Tracker {Core.Version} @ Hearthsim";
+			var json = JObject.Parse(wc.DownloadString("https://api.github.com/rate_limit"));
+			wc.Dispose();
+			if(json["rate"]["remaining"].ToString() == "0")
+				return true;
+			return false;
+		}
+
+		private string releaseUrl(string repoUrl)
```
azeier: Methods should be CamelCase

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+
+		#region Available
+
+		private void ButtonInstall_OnClick(object sender, RoutedEventArgs e)
+		{
+			try
+			{
+				var wc = new WebClient();
+				wc.Headers["User-Agent"] = $"Hearthstone Deck Tracker {Core.Version} @ Hearthsim";
+				var releaseUrl = (ListBoxAvailable.SelectedItem as Plugin).ReleaseUrl;
```
azeier: This should either be a cast or have a null check

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+				var json = JObject.Parse(wc.DownloadString(releaseUrl));
+				var downloadUrl = json["assets"].First["browser_download_url"].ToString();
+				var downloadFile = Path.Combine(Path.GetTempPath(), downloadUrl.Split('/').Last());
+				wc.DownloadFile(downloadUrl, downloadFile);
+				var stringList = new string[] { downloadFile };
```
azeier: No need for `string` here

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+				Log.Error(ex);
+				GithubUnavailable().Forget();
+			}
+		}
+
+		
+		private void ButtonUninstall_OnClick(object sender, RoutedEventArgs e)
+		{
+			var plugin = ListBoxPlugins.SelectedItem as PluginWrapper;
+			var parent = Directory.GetParent(plugin.RelativeFilePath);
```
azeier: `plugin` can be null , should be either a cast or needs a null check

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+			Helper.TryOpenUrl(releaseUrl);
+		}
+
+
+		#endregion
+
+		private void ButtonUpdate_OnClick(object sender, RoutedEventArgs e)
+		{
+			try
+			{
```
azeier: `using`, object intializer, cast/nullcheck `SelectedItem`

Hearthstone Deck Tracker/Plugins/PluginManager.cs
```diff
@@ -200,6 +200,12 @@ private IEnumerable<PluginWrapper> GetModule(string pFileName, Type pTypeInterfa
 						if(typeInterface == null)
 							continue;
 						var instance = Activator.CreateInstance(type) as IPlugin;
+						var updatable = Activator.CreateInstance(type) as IUpdatable;
+						if(updatable != null)
+						{
+							plugins.Add(new PluginWrapper(pFileName, instance, updatable));
+							continue;
```
azeier: This can just be an `else` on the `if` below.

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerPlugins.xaml.cs
```diff
@@ -65,52 +70,319 @@ private void ButtonOpenPluginsFolder_OnClick(object sender, RoutedEventArgs e)
...
+				Log.Info("Checking for updates to installed plugins.");
+				var updateList = new List<Plugin>();
+				if(ListBoxPlugins.Items.IsEmpty)
+					return null;
+				var wc = new WebClient();
+				foreach(var p in ListBoxPlugins.Items)
+				{
+					var plugin = p as PluginWrapper;
+					wc.Headers["User-Agent"] = $"Hearthstone Deck Tracker {Core.Version} @ Hearthsim";
```
judge2020: The header is inside the loop because when declared outside the loop, it would be reset to null after the first iteration and GitHub would not serve the request.

judge2020:
I think it's ready :) I'm open to any more suggestions or code I should refine.

It seems to work good right now, I tested running when the user has hit Github's rate limit and it seems to work fine. If so, when downloading available, it uses a cached version.

If it hits when checking for updates, it will simply shop "up to date" and will only notify the user they have hit the rate limit if they try to update and have hit it. 

One thing we could do to fix the GitHub rate limit issue may be to include a GitHub API key in the config, however, it may get hit too quick as more people are likely to use plugins over the Imgur screenshot uploader, which was already hitting the rate limit some (#2806).

Another solution may be to obtain via a cache on HearthSim's and/or host a list of plugins and update details that would be updated every day/*x* hours. This would mean it only requests from github when a user goes to download, and not when checking.

azeier:
Starting to look really good for the most part! :)

Haven't looked at the code yet but here are some UI/UX changes I would like to see:

### Installed
- Updating plugins freezes the app
- "Updating" and "Update installed" still look like hyperlinks
- "Update installed" is clickable and causes the update to run again
- "Update installed" should probably say "Restart required" and restart the app when clicked (Maybe "Installed. Restart now"? not entirely sure on wording, something like that though)
- "Up to date [checkmark]" should be gray/green, blue shouldn't be used for non-hyperlinks
- Newly installed plugins dont say "up to date"
- "Details" is disabled on newly installed plugins
- Uninstalling a plugin leaves it with no plugin selected and empty, non-functional buttons
- "Details" should probably be renamed "Project"



### Available
- Installing a plugin freezes the app
- Installing a plugin doesn't remove it from the list
- Installing a plugin should move you to the "installed" page with the plugin selected (this is debatable but I think correct)
- Clicking "Details" opens an API page
- "Details" should probably be renamed "Project"
- "Details" and the plugin button should be swapped: [plugin][project][uninstall]
- Separator below the buttons needs some top margin
- "Details" and "Install" are clickable without selecting a plugin (bottom section should either be blank or a plugin needs to be selected by default)
  - Clicking "Details" crashes the app
  - Clicking "Install" does something and locks the install button

 
### More general
- Plugin names/descriptions in are not always consistent with the installed versions
- Rate limit message needs adjustment, Users probably don't know what github is and "Try again in 2263 or manually drag-and-drop the plugin." doesn't explain things too well.
  - "Too many requests", "Please try again in {x} minutes or manually download and install the plugin.\nGo to the download page now?"
  - Buttons: "Go to download", "Try again later"

- I see a bunch of these message in the console: 
```System.Windows.Data Error: 6 : 'SystemConvertConverter' converter failed to convert value '<null>' (type '<null>'); fallback value will be used, if available. BindingExpression:Path=UpdateTextEnabled; DataItem='PluginWrapper' (HashCode=3372975); target element is 'Hyperlink' (HashCode=37657477); target property is 'IsEnabled' (type 'Boolean') InvalidCastException:'System.InvalidCastException: Null object cannot be converted to a value type.
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at MS.Internal.Data.SystemConvertConverter.Convert(Object o, Type type, Object parameter, CultureInfo culture)
   at System.Windows.Data.BindingExpression.ConvertHelper(IValueConverter converter, Object value, Type targetType, Object parameter, CultureInfo culture)'
```

judge2020:
I'll probably revisit this in a while when I have more time and a clear vision for the end goal, instead of something like this where it becomes something more and the code style falls apart.

PR #3120, opened 2017-03-08
Include importing for DisguisedToast.com

judge2020:
Includes importing support for DisguisedToast.com decks and adds a unit test.

The unit test includes many cards due to there being no older decks on there that are similar to Patron/charge warrior.


jleclanche:
I'll talk to him, see if we can add native HDT imports

judge2020:
That would be best, to get meta tags on the website.

PR #3135, opened 2017-03-28
Recolor Hyperlink colors into Metro light blue 

judge2020:
The current hyperlinks are a little dark for both dark and light themes for HDT. This will recolor them for the default accent's light blue (although they work nicely with other accents).

### DIFF
--- a/Hearthstone Deck Tracker/App.xaml
+++ b/Hearthstone Deck Tracker/App.xaml
@@ -26,6 +26,11 @@
                 <ResourceDictionary>
                     <utility:ConfigWrapper x:Key="ConfigWrapper"/>
                 </ResourceDictionary>
+                <ResourceDictionary>
+                    <Style TargetType="{x:Type Hyperlink}">
+                        <Setter Property="Foreground" Value="#FF47B1DF" />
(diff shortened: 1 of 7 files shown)


azeier:
There's a better way to achieve this.

First way (better, but still bad if it's used app/control wide):
Add `<Color x:Key="HyperlinkColor">#FF47B1DF</Color>` to resource dictionary in `App.xaml`, then use `Foreground="{StaticResource HyperlinkColor}"`

Best way if used consistently:
Setting this in `OptionsMain.xaml`
```
<UserControl.Resources>
    <Style TargetType="{x:Type Hyperlink}">
        <Setter Property="Foreground" Value="#FF47B1DF" />
    </Style>
</UserControl.Resources>
```

judge2020:
If any new hyperlinks were to be put in and are not desired to be this color, it could be changed from option 2 to 1 easily. I'm not sure which one is most appropriate in this case.

judge2020:
I think using the second solution works. A hyperlink color can be overridden by explicitly setting it, as with the news bar at the top of the screen which is still white in this instance,

PR #3177, opened 2017-04-09
Changed from bootstrap.bat to git submodules for dependencies.

distaula:
I setup the external dependencies that were in the bootstrap.bat file and made them git submodules.  This can allow the app to target certain commits if needed and prevents the need from running any sripts before building the app in Visual Studio.

Most git UIs handle the modules automatically now, or you can use --recursive in the command-line and be able to build the app immediately after cloning.

I also moved copying of the localization resources from the .bat script to the .csproj so the project can build without it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,11 +159,7 @@ packages/*
 !packages/repositories.config
 
 Hearthstone Deck Tracker/Releases
-HearthDb/
-HearthMirror/
 Hearthstone Deck Tracker/Resources/CHANGELOG.md
-HSReplay-Api/
 Hearthstone Deck Tracker/Properties/Strings.*
-HDT-Localization/
 
 Resources/Generated/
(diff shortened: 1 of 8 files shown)


jleclanche:
Doing this will require constantly updating the submodule reference though, which is a PITA especially for the i18n repo. If anything they should be subtrees?

distaula:
It should default to the module's master/head unless something else is explicitly specified, which is what the app was doing with the bootstrap script, I haven't needed to update any references.  Am I missing something?  (I definitely could be)

The failed check is just because I've removed the dependency on bootstrap.bat, and it doesn't look like the AppVeyor.yml is part of the repo.

jleclanche:
It defaults to the head when initializing, but your commit initializes it. Submodules always have a commit reference in a tree. Thus if this gets merged as-is, it will forever reference the commits it initialized to, until it's updated.

jleclanche:
In other words, submodules work nicely when you are dealing with libraries which are infrequently released. Not if you're dealing with a cousin repo.

azeier:
I have nothing to add to that. No good reason to do this.

PR #3217, opened 2017-04-19
Fixed Pre-Build Event Issues and Bootstrap / Resource Gen Errors

jbakert:
Several batch files would throw errors in the original solution and cause problems when the HSDT projects were included in a 3rd party solution (usually a plugin). 

Added error checking to scripts to prevent execution of routines that were contingent upon a successful completion of the previous command.  

Added the ability to disable the PreBuildEvent if you so desired. Fixed missing nuget.exe for VS2017 by using msbuild's new integrated functionality. 

Also rocked a nice ^M in there somewhere.

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
+++ b/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
@@ -1200,6 +1200,7 @@
     </Compile>
     <Compile Include="Utility\Updating\Updater.Squirrel.cs" />
     <Compile Include="Utility\Updating\Updater.Default.cs" />
+    <None Include="Test.bat" />
     <Resource Include="Resources\hsreplay_logo_white.png" />
     <EmbeddedResource Include="Resources\CHANGELOG.md" />
     <EmbeddedResource Include="Properties\Strings.ja-JP.resx">
@@ -1315,6 +1316,7 @@
     <Resource Include="Resources\ClassIcons\General\BaseLight\all.png" />
     <Resource Include="Resources\ClassIcons\General\BaseLight\archived.png" />
     <Resource Include="Resources\ClassIcons\General\BaseDark\archived.png" />
+    <None Include="PreBuildEvent.bat" />
     <None Include="Resources\card-mark-coin.png" />
     <Resource Include="Utility\Markdown.Xaml\License.txt" />
     <Resource Include="Resources\cards.png" />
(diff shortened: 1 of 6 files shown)


Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
```diff
@@ -1398,36 +1400,9 @@
...
-) else (
-  git clone --depth 1 https://github.com/HearthSim/HDT-Localization.git $(SolutionDir)HDT-Localization
-)
-xcopy /Y "$(SolutionDir)HDT-Localization\*.resx" "$(ProjectDir)Properties\"
-mkdir "$(TargetDir)\Images\Tiles"
-mkdir "$(TargetDir)\Images\Themes"
-xcopy /E /Y /Q "$(SolutionDir)Resources\Generated\Tiles" "$(TargetDir)\Images\Tiles"
-xcopy /E /Y /Q "$(ProjectDir)Images\Themes" "$(TargetDir)\Images\Themes"
-xcopy /Y /Q "$(SolutionDir)CHANGELOG.md" "$(ProjectDir)Resources"</PreBuildEvent>
+    <PreBuildEvent>"$(ProjectDir)PreBuildEvent.bat" "$(ProjectDir)..\" "$(ProjectDir)" "$(ProjectDir)($OutDir)" 0 0
```
azeier: This should probably be `$(OutDir)`? Also, why not keep `$(TargetDir)`?

bootstrap.bat
```diff
@@ -1,4 +1,7 @@
-nuget restore
+@echo off
+
+msbuild /t:restore
```
azeier: This doesn't seem to work for me. I get the same error as [AppVeyor](https://ci.appveyor.com/project/azeier/hearthstone-deck-tracker) when running bootstrap locally on a clean clone.

bootstrap.bat
```diff
@@ -1,4 +1,7 @@
-nuget restore
+@echo off
+
+msbuild /t:restore
+REM nuget restore
```
azeier: Assuming the line above works, this can probably be removed.

Hearthstone Deck Tracker/Test.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
```
azeier: This file should probably be removed.

bootstrap.bat
jbakert: I noticed a lot of nuget references in your output. Is it possible that the version of nuget you're using is incompatible? When looking for a nuget for 2017, I came across an article that stated the command line nuget was deprecated and being phased out by being integrated into msbuild

bootstrap.bat
jbakert: The REM?  Yes, I agree.

Hearthstone Deck Tracker/Test.bat
jbakert: Yes

bootstrap.bat
jbakert: Welp. I was playing with .NET Core. I've reverted the change. Would it be prudent to include the nuget 4.1 binary in the root solution folder?

judge2020: It's usually recommended to have Nuget in your path somewhere.

Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
jbakert: Yes, I made this change.

bootstrap.bat
jbakert: Fair. But path parsing order, multiple scattered versions, and explicit binaries necessary to get a solution in working order mimic the package system, so I think the addition of this tool to be prudent. That being said, I don't mind either way.

azeier:
Not a fan of the submodules or including nuget.exe. Otherwise this looks gtm.

judge2020:
^ 
Most of the time, it is recommended to have nuget somewhere (else) in your path. It's easy to download it from nuget.org anyways.

PR #3257, opened 2017-05-24
Current Tutorial leads to build errors.  This line fixes that

HartnettMichael:
When I used the tutorial I had Hearthstone Deck Tracker.exe not found and resgen.exe not found.  This fixes that.  If this pull request is rejected, then could you change the tutorial to add what worked for me.  Go to HearthstoneDeckTRacker and unload project.  Edit HearthStone Deck Tracker .csproj.  Go twoards the bottom where you will find Target GenerateDesignerFiles.  Add SdkToolsPath="$(TargetFrameworkSDKToolsDirectory)"
after the following Generate Resource.  It will look like this.
```
<Target Name="GenerateDesignerFiles">
    <GenerateResource SdkToolsPath="$(TargetFrameworkSDKToolsDirectory)" Sources="Properties/Strings.resx" StronglyTypedLanguage="C#" StronglyTypedClassName="Strings" StronglyTypedNamespace="Hearthstone_Deck_Tracker.Properties" StronglyTypedFileName="Properties/Strings.Designer.cs" OutputResources="Hearthstone_Deck_Tracker.Properties.Strings.resources" PublicClass="true">
    </GenerateResource>
    <Delete Files="Hearthstone_Deck_Tracker.Properties.Strings.resources" />
  </Target>
  <Target Name="BeforeBuild">
    <CallTarget Targets="GenerateDesignerFiles" />
  </Target>
```

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
+++ b/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
@@ -1479,7 +1479,7 @@
   </Target>
   -->
   <Target Name="GenerateDesignerFiles">
-    <GenerateResource Sources="Properties/Strings.resx" StronglyTypedLanguage="C#" StronglyTypedClassName="Strings" StronglyTypedNamespace="Hearthstone_Deck_Tracker.Properties" StronglyTypedFileName="Properties/Strings.Designer.cs" OutputResources="Hearthstone_Deck_Tracker.Properties.Strings.resources" PublicClass="true">
+    <GenerateResource SdkToolsPath="$(TargetFrameworkSDKToolsDirectory)" Sources="Properties/Strings.resx" StronglyTypedLanguage="C#" StronglyTypedClassName="Strings" StronglyTypedNamespace="Hearthstone_Deck_Tracker.Properties" StronglyTypedFileName="Properties/Strings.Designer.cs" OutputResources="Hearthstone_Deck_Tracker.Properties.Strings.resources" PublicClass="true">
     </GenerateResource>
     <Delete Files="Hearthstone_Deck_Tracker.Properties.Strings.resources" />
   </Target>


azeier:
If it doesn't do any harm I'm not really against merging it, but do you know why it's needed in your case? I don't seem to need it and I haven't heard of anyone else having this issue so far.

riQQ:
I've got the same problem.
Error message is the following on Windows 7 x64:
>error MSB3091: Task failed because "resgen.exe" was not found, or the correct Microsoft Windows SDK is not installed. The task is looking for "resgen.exe" in the "bin" subdirectory beneath the location specified in the InstallationFolder value of the registry key HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\Windows\v8.0A\WinSDK-NetFx35Tools-x86. You may be able to solve the problem by doing one of the following:  1) Install the Microsoft Windows SDK.  2) Install Visual Studio 2010.  3) Manually set the above registry key to the correct location.  4) Pass the correct location into the "ToolPath" parameter of the task.

The windows sdk is not installed. The registry key does not exist on my pc.

riQQ:
After installing the `.NET Framework 4.5 SDK` from the `Windows Software Development Kit (SDK) for Windows 8`, the error stays the same. I don't know why it's looking for the NetFx 3.5 Tools. The Windows SDK 8 only contains NetFx 4.0 Tools.

riQQ:
With the proposed change, `SdkToolsPath=C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\` is used. This seems to be wrong, too.

omni5cience:
I also had this issue, and this fix solved the problem for me. I didn't dig too far into the cause of the issue, but FWIW this got me unblocked.

riQQ:
I tested this with a simple test project with resources and targeting .NET Framework 4.5:
The `GenerateResources` task is called with `Task Parameter:SdkToolsPath=C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.6.1 Tools\`, so this change should be correct.

@azeier could you check which path is used on your machine by setting the `MSBuild project build output verbosity` to `Diagnostic` in `Tools > Options > Project and Solutions > Build and Run`.

riQQ:
VisualStudio calls `GenerateResource` with `$(TargetFrameworkSDKToolsDirectory)`, if `$(ResgenToolPath)` isn't defined, which it isn't by default (see `Microsoft.Common.CurrentVersion.targets`). So in my opinion this PR is correct.

PR #3258, opened 2017-05-24
Fixes Mirror Entity edge case. 

HartnettMichael:
https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/3253
Mirror entity will not gray out because player plays a minion and their opponent has 7 minions.


azeier:
Please use tabs for indenting.

PR #3276, opened 2017-06-04
Detail error loading plugin

TheBestForFun:


Hearthstone Deck Tracker/Plugins/PluginManager.cs
```diff
@@ -214,7 +214,10 @@ private IEnumerable<PluginWrapper> GetModule(string pFileName, Type pTypeInterfa
 			}
 			catch(Exception ex)
 			{
-				Log.Error("Error loading " + pFileName + ":\n" + ex);
+				var reason = (ex is System.Reflection.ReflectionTypeLoadException)
+					? String.Join("\n", (ex as ReflectionTypeLoadException).LoaderExceptions.Select(exp => exp.Message))
```
azeier: Even though this is no longer necessary when it's a separate case: `as` implies the value could be null, so `(ex as ReflectionTypeLoadException)` should either be a cast or handle the null case. And since there is a type check before it, this should be a cast.

azeier: Also please use the `string` keyword rather than the `String` class.

Issue #3296, opened 2017-06-14
Include deck version number when exporting to HS via code


garyking:
In the new HDT, when we export a deck, we're given code to copy instead. Please include the deck's version number in the deck's title, like it was before.

PR #3297, opened 2017-06-15
Include deck version in export dialog

judge2020:
Addresses #3296. 
img

Added the HDT version as well, will be happy to change/remove it if necessary.


azeier:
This should not be the default.

I don't think the HDT version needs to be included.

judge2020:
img2

img3

Hearthstone Deck Tracker/Config.cs
```diff
@@ -341,6 +341,9 @@ public class Config
 		[DefaultValue(0.13)]
 		public double ExportClearY = 0.13;
 
+		[DefaultValue(true)]
+		public bool ExportIncludeVersion = true;
```
azeier: Default should be changed to false, then this is gtm.

PR #3301, opened 2017-06-22
Fixes the ambiguity of the date in Arena and Brawl decks

FreddyFlares:
What do you think about this change? Some of the issues mention it and it was annoying to see a deck named 'ARENA 07-06 03:15' when you played it at 15:15 (in the afternoon).

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -293,8 +293,8 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue(true)]
 		public bool EnterToSaveNote = true;
 
-		[DefaultValue(true)]
-		public bool ExportIncludeVersion = true;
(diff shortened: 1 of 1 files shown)


azeier:
Looks good. Please remove the first commit though.

azeier:
Also just fyi: the email you used to commit is not associated with your github account.

FreddyFlares:
Added the email as primary, working on the rest, still learning git. Had to jump through a few hoops to get bootstrap.bat to complete on my setup.

One thing to ask: bootstrap appears to clone separate repos as submodules. The VS 2017 Github extension sees these as changes to commit which I can't get rid of, even though the folders are in .gitignore, but Git GUI doesn't. Research is telling me that 'gitlinks' are created to represent submodules and I guess that's what my commit above did.

Anyway for now I will commit with the GUI so the submodules aren't committed.

PR #3304, opened 2017-06-23
Size of Card-marks

Incintt:
Hey!

I usually play HS in windowed mode (small size). 

In order to use HS DeckTracker I change the scaling of Player\Opponent to 40. 
However, card-marks size doesn't change when i change scaling of player\opponent.
It looks like this:
image

I look forward when it will be fixed in future patches. 

Thanks!

Best regards,
Incin


azeier:
Please [open an issue for that](https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/new), not a pull request.

PR #3329, opened 2017-07-21
Don't show global statistics for arena games

### LINKED ISSUE #3328: Don't show global statistics for arena games

riQQ:
Closes #3328


azeier:
This has inconsistent behavior between switching from an arena or a non-arena deck to "no deck mode". It should probably show the panel in either case.

PR #3334, opened 2017-07-23
Fix HearthstoneTopDecks import

mgiuffrida:
This fixes the `HearthstoneTopDecks` test. Also tested manually in HDT.


azeier:
I'll merge this since it's a very minor change, but in general we'd rather encourage websites to implement [meta tag based importing](https://github.com/HearthSim/Hearthstone-Deck-Tracker/wiki/RFC-Importing-Implementation).

mgiuffrida:
understood, I have no affiliation with the site -- thanks though!

Issue #3344, opened 2017-07-30
Include non-checkbox options in options search results


mgiuffrida:
### Expected Behavior
Search results in Options => Search should include check boxes, but also sliders, dropdown menus, text boxes, and possibly buttons.

### Actual Behavior
Only check boxes are included.

azeier:
Shouldn't be too though. [This](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs) needs to be extended to support any type of FrameworkElement and have a full list of "viable types".

Feel free to take a crack at it! :)

PR #3346, opened 2017-07-30
Refactor OptionsSearch ahead of supporting more option types

mgiuffrida:
Instead of hard-coding `CheckBox` as the type of option to search for, use a
base class `OptionWrapper`. Currently `CheckBoxWrapper` is the only concrete
child class, but more will be added for #3344.

Use a filter delegate to check whether an element is an "option" so we can
explicitly filter for `CheckBox` types, and for `DockPanel`s that contain a
`CheckBox`. In the future, `DockPanel`s that contain other types of controls
will also be selected. Finding the `DockPanel`s will let us find the related
text depending on the control type, as non-CheckBox types tend to have more
complicated UI.

This change should not affect search results today.

### DIFF
--- a/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
+++ b/Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
@@ -1,5 +1,6 @@
 ﻿#region
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows;
@@ -15,7 +16,7 @@ namespace Hearthstone_Deck_Tracker.FlyoutControls.Options
 	/// </summary>
 	public partial class OptionsSearch : UserControl
 	{
-		private List<CheckBoxWrapper> _checkBoxWrappers;
+		private List<IOptionWrapper> _optionWrappers;
 
 		public OptionsSearch()
 		{
@@ -27,13 +28,13 @@ namespace Hearthstone_Deck_Tracker.FlyoutControls.Options
 			((TextBox) sender).Focus();
 		}
 
-		private List<CheckBoxWrapper> CheckBoxWrappers => _checkBoxWrappers ?? (_checkBoxWrappers = LoadWrappers());
+		private List<IOptionWrapper> OptionWrappers => _optionWrappers ?? (_optionWrappers = LoadWrappers());
 
 		private void ButtonSearch_OnClick(object sender, RoutedEventArgs e) => UpdateSearchResult(TextBoxSearch.Text);
 
-		private List<CheckBoxWrapper> LoadWrappers()
+		private List<IOptionWrapper> LoadWrappers()
 		{
-			var options = new[]
+			var optionsMenuItems = new[]
 			{
 				new UserControlWrapper(Core.MainWindow.Options.OptionsOverlayDeckWindows, nameof(Core.MainWindow.Options.OptionsOverlayDeckWindows)),
 				new UserControlWrapper(Core.MainWindow.Options.OptionsOverlayGeneral, nameof(Core.MainWindow.Options.OptionsOverlayGeneral)),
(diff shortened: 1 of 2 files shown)


Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
 			{
-				if(selected.CheckBox.Visibility == Visibility.Collapsed)
+				if(selected.OptionControl.Visibility == Visibility.Collapsed)
 					AdvancedOptions.Instance.Show = true;
 				target.IsSelected = true;
 			}
 		}
 
-		public class CheckBoxWrapper
+		public abstract class OptionWrapper
```
mgiuffrida: I'd prefer to make this a generic `public abstract class OptionWrapper<T>` because we really want to parameterize `OptionControl`. For instance, `CheckBoxWrapper` would be an `OptionWrapper<CheckBox>` whose `OptionControl` is already a `CheckBox`, so it wouldn't need a separate `CheckBox` property that refers to the same control as `OptionControl`.

But I couldn't make that work with collections. eg, I couldn't find a way to do this:

```
private List<OptionWrapper<T>> LoadWrappers()
{
  var optionWrappers = new List<OptionWrapper<T>>; 
  // ...  
}
```

Any thoughts? I haven't touched C# in 9 years...

mgiuffrida: Actually I found a way to do this by making an interface. [Here's what that looks like diffed against this PR.](https://github.com/mgiuffrida/Hearthstone-Deck-Tracker/commit/search-more-options-generic) But maybe that's more complicated than it's worth? Deriving an `OptionWrapper<T>` from an `interface OptionWrapper` is weird to me, coming from C++.

Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
+				optionWrappers.AddRange(option.Select(cb => WrapOption(optionsMenuItem, cb)));
 			}
-			return checkBoxWrappers;
+			return optionWrappers;
+		}
+
+		private bool IsSearchableOption(DependencyObject depObj)
+		{
+			var basicTypes = new List<Type>{typeof(CheckBox)};
+			if(basicTypes.Contains(depObj.GetType())) return true;
```
mgiuffrida: Done.

mgiuffrida: Done

Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
+			if(depObj is DockPanel)
```
mgiuffrida: Sweet. Done in a few places.

Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
+						return WrapSimpleOption(menuItem, child as CheckBox);
+				}
+			}
+			return WrapSimpleOption(menuItem, depObj as ContentControl);
+		}
+
+		private OptionWrapper WrapSimpleOption(UserControlWrapper menuItem, ContentControl control)
+		{
+			if(control == null)
+				throw new ArgumentNullException("control");
```
mgiuffrida: Done.

Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
 				target.IsSelected = true;
 			}
 		}
 
-		public class CheckBoxWrapper
+		public abstract class OptionWrapper
 		{
-			public CheckBox CheckBox { get; set; }
-			public UserControlWrapper MenuItem { get; set; }
+			public OptionWrapper(UserControlWrapper menuItem)
```
mgiuffrida: Done.

Hearthstone Deck Tracker/FlyoutControls/Options/OptionsSearch.xaml.cs
```diff
@@ -50,44 +51,111 @@ private List<CheckBoxWrapper> LoadWrappers()
...
+
+		public class CheckBoxWrapper : OptionWrapper
+		{
+			public CheckBoxWrapper(UserControlWrapper menuItem, CheckBox checkBox) : base(menuItem)
+			{
+				CheckBox = checkBox;
+			}
+
+			private CheckBox CheckBox { get; }
+			public override Control OptionControl { get => CheckBox; }
```
mgiuffrida: Done.

mgiuffrida:
Thanks for the comments. I went with the generic base class implementing an interface. If you'd like to see the changes  specific to your comments, see https://github.com/mgiuffrida/Hearthstone-Deck-Tracker/commit/4075c8b6d16773a5e5d3c3118853888865c143a8.

azeier:
Perfect 👍

PR #3382, opened 2017-08-19
Add custom import for Hearthhead decks

danblair:
Hearthhead no longer supplies meta tags for deck import. This change would use the Hearthhead JSON data that is normally loaded into the windows space of the browser to discover the deck name and deck string for import.

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
+++ b/Hearthstone Deck Tracker/Hearthstone Deck Tracker.csproj
@@ -264,6 +264,7 @@
     </Compile>
     <Compile Include="FlyoutControls\DeckExport\DeckExportViewModel.cs" />
     <Compile Include="HsReplay\Data\JsonCache.cs" />
+    <Compile Include="Importing\Websites\Hearthhead.cs" />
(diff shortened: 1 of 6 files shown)


azeier:
[We no longer want to maintain custom importers](https://github.com/HearthSim/Hearthstone-Deck-Tracker/wiki/Web-importing#legacy-support).

This is something they should fix on their end.

As a workaround: You can use their copy deck button for importing decks to HDT.

danblair:
I can certainly agree with not creating a brittle import.  What about removing the Hearthhead unit test then?

edzam:
Looks like I when I migrated Hearthhead to our new framework some things got missed.  This is one of them.  Ill get this ASAP.

judge2020:
@edzam if you'd like, the meta tags were upgraded to where it needs just the deck string over the individual cards https://github.com/HearthSim/Hearthstone-Deck-Tracker/wiki/RFC-Importing-Implementation.

azeier:
Thanks @edzam :).

The test now works again - might as well keep it. The "Web" tests are not run by CI anyway.

PR #3485, opened 2017-12-03
Add hidden cache secret tracking logic

### LINKED ISSUE #2987: "Hidden Cache" secret not greying out
If the opponent is Hunter and plays a secret, then after I play a minion, "Hidden Cache" should grey out from the list of secrets.

riQQ:
Resolves #2987.

### DIFF
--- a/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
+++ b/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
@@ -32,7 +32,8 @@ namespace HDTTests.Hearthstone.Secrets
 			_secretMage2,
 			_secretPaladin1,
 			_secretPaladin2,
-			_opponentEntity;
+			_opponentEntity,
+			_opponentCardInHand1;
 
 		private Entity CreateNewEntity(string cardId)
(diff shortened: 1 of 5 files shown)


azeier:
This doesn't actually fix #2987 though, does it? This fixes the edge case where the opponent has a revealed minion in hand, i.e. one that was returned from play - which means something like sap or freezing trap was played by the player. In the majority of cases it will still not grey out correctly because it's either not possible to know or requires card type deduction based on played cards.

riQQ:
No, as an opponent plays cards their card type is revealed and if it's a minion, the secret is greyed out. It worked in my test.

azeier:
If your opponent has not yet played any cards but one secret and you play a minion. Nothing happens. You don't know whether that's because he has no minion in hand or it's not hidden cache, because none of the cards in his hand have the card type revealed.

azeier:
Oh - I see, this handles that after the fact, interesting.

riQQ:
Yes, I know. But in this case all cards in the opponent's hand are tracked and checked when the data of the corresponding entity is updated (e.g. through playing or through a reveal).

PR #3570, opened 2018-02-20
Resource Generator v2

Latency:
See https://github.com/HearthSim/Hearthstone-Deck-Tracker/pull/3538


azeier:
This seems unlikely to happen and is not mergable at this point anymore anyway.

Issue #3569, opened 2018-02-19
Secret tracker uses Standard secrets cardpool during Wildfest Arena


alitch:
### Expected Behavior
In arena games for the Wildfest event, the secret tracker should be using the wild cardpool.

### Actual Behavior
Secret tracker only lists and checks for secrets in the current standard cardpool.

### Steps to reproduce behavior
Enter a Wildfest Arena game and have the opponent play a secret.

PR #3574, opened 2018-03-01
Remove Standard Arena Secret filter for Wild Arena Fest

alexkrohg64:
Removed the Standard filter for Arena secrets. This will need to be reverted when Wild Arena ends.

Resolves: Issue [#3569](https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/3569)


alexkrohg64:
Looks to be working on my build.

wildarenasecrets

azeier:
Didn't get around to this, sorry.

Might look into a more generic solution if this event happens again.

PR #3592, opened 2018-03-14
Twitter icon color fix

kadavr95:
Twitter icon was different color from other icons in the menu. Not sure if it was intended difference to highlight it, so I had decided to fix it.

Current master:
image
After change:
image


azeier:
That was indeed not intended. 

We require singled CLAs from contributors nowadays (I really should put that somewhere), since this is a very minor non-code fix I'm just gonna push a commit myself so we don't have to do that here 😄.

We're always happy about HDT contributors though, if you're generally interested just shoot me a message here or on discord!

PR #3595, opened 2018-03-15
change language of comboboxes' items in option when change language

minn951120:


minn951120:
In last version,
if I change language in option, items of comboboxes were not apply that change.
so I corrected it by reloading options and refreshing comboboxes.

minn951120:
Reloading is not necessary, so I'll take another request

PR #3668, opened 2018-05-05
Add setting to allow closing to tray

xeropresence:

I often hit close but what I really wanted was just to close the window, not to terminate the application. Add a setting that when the close button is hit it will instead minimize. 

I added a public variable to the form that the tray icon can set, I thought this might be better then adding an entire new function.

Unsure about how to handle the localization of Options_Tracker_Settings_CheckBox_CloseToTray


azeier:
Looks good. I can take care of the localization for it.

Just fyi: looks like the email you used to commit isn't associated with your github account.

PR #3681, opened 2018-05-13
make cards created by Shadow Reflection work correctly

### LINKED ISSUE #3536: Shadow Reflection + Jade Idol (bug)
### Expected Behavior
After selecting to add 3 Jade Idols (with Shadow Reflection) it should add 3 Idols to card list

### Actual Behavior
It doesn't add anything. Probably doesn't work with any power cards.

minn951120:
fixed #3536 
make cards created by Shadow Reflection (not only Jade Idol) work correctly


azeier:
Sorry for ignoring the previous one, meant to comment that something like this should be possible. Much nicer solution indeed, thanks!

PR #3686, opened 2018-05-17
Added tracking for several Monster Hunt cards

Mikeywalsh:
Added tracking for the following cards:
- Creepy Curio
- Haunted Curio
- Old Militia Horn
- Militia Horn
- Surly Mob
- Angry Mob


azeier:
Thanks :)

PR #3687, opened 2018-05-18
HS Alerts (Flash & Popup)

### LINKED ISSUE #3685: New feature idea for alerts
Hey, i am curious if this is possible from a technical point and if it is, if there is a practical reason it should not be implemented.

Today we have the option to make hearthstone popup whenever your turn starts.
We also have the option to make hearthstone flash when you get a incoming friendly challenge.

Question 1: Is it technically possible to make hearthstone popup whenever you get a friendly challenge?

Question 2: Is it technically possible to make hearthstone popup whenever you recieve a message from a friend? 

Many people ask you if you want to play 80G quest, and if you don't answer, they will ask the next friend on their list instead of challenging you.
Because the way it works now, i always miss the flashing alert in HDT because that window is minimized (browsing the web/watching youtube)

If this is not technically possible/or this is a feature you don't want to implement because it would put more load on blizzards servers (more inactive players hanging out in main menu) could you atleast add the option for a SOUND alert in HDT? 

The flashing alert is not good enough, since it's so easy to miss.
I would LOVE to get the popup option for friendly challenges and incoming messages.
But if not, PLEASE add the option of a SOUND alert in HDT.

Really appreciate all your hard work, HDT is an amazing tool :)

CLJunge:

Resolves #3685, moved the Alerts section from General to Notifications. I've also added a new key (Options_Tracker_General_CheckBox_Popup_Challenge), but i don't how to add the translations using crowdin.

### DIFF
--- a/Hearthstone Deck Tracker/Config.cs
+++ b/Hearthstone Deck Tracker/Config.cs
@@ -143,8 +143,11 @@ namespace Hearthstone_Deck_Tracker
 		[DefaultValue("Dungeon Run {Date dd-MM HH:mm}")]
 		public string DungeonRunDeckNameTemplate = "Dungeon Run {Date dd-MM HH:mm}";
 
-		[DefaultValue(false)]
-		public bool BringHsToForeground = false;
+		[DefaultValue(HsActionType.Flash)]
+		public HsActionType? TurnAction = HsActionType.Flash;
+
+		[DefaultValue(HsActionType.Flash)]
+		public HsActionType? ChallengeAction = HsActionType.Flash;
 
 		[DefaultValue(false)]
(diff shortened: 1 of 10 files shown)


CLJunge:
I implemented the changes you requested, but the translations are still missing. How do I add new keys, do I simply just checkout the HDT-Localization project and edit the Strings.resx file?

azeier:
Sorry for the delay here. Changes look good. 

I added the new strings for you but yes, that is exactly how you would do it.
For reference: https://github.com/HearthSim/HDT-Localization/commit/2407788f9d1cddc8188b84d7168bbc6f497feb8c

PR #3710, opened 2018-06-24
[Options] Only use ScrollViewer where needed

CLJunge:

The buttons at the bottom of Options/Tracker/Plugins are hidden if the user has multiple plugins installed. Can be confusing and annoying, so I only added a ScrollViewer control where it's actually needed.


azeier:
This doesn't seem like the right solution. I shouldn't have to check whether an options panel needs to now have a scrollviewer every time I add something new to it.

CLJunge:
Yeah, you're right, I'll look for a different solution.

PR #3800, opened 2018-10-08
Fix Dungeon Run deck name

daltoncwu:
Fix Dungeon Run deck name incorrectly being called "Monster Hunt"
Fixes https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/3697

### DIFF
--- a/Hearthstone Deck Tracker/Hearthstone/DungeonRun.cs
+++ b/Hearthstone Deck Tracker/Hearthstone/DungeonRun.cs
@@ -23,6 +23,10 @@ namespace Hearthstone_Deck_Tracker.Hearthstone
 				LastEdited = DateTime.Now
 			};
 			var template = GetDeckTemplate(set);
+			if(template is null)
+			{
+				return null;
+			}
 			deck.Name = Helper.ParseDeckNameTemplate(template, deck);
 			return deck;
 		}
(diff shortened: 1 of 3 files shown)


Hearthstone Deck Tracker/Hearthstone/DungeonRun.cs
```diff
@@ -22,12 +22,24 @@ public static Deck GetDefaultDeck(string playerClass, CardSet set)
 				IsDungeonDeck = true,
 				LastEdited = DateTime.Now
 			};
-			//Hack to avoid multiple templates. Should be good enough 99.9% of the time.
-			var template = Config.Instance.DungeonRunDeckNameTemplate.Replace("Dungeon Run", "Monster Hunt");
+			var template = getDeckTemplate(set);
```
azeier: This needs to more gracefully handle the case where `GetDeckTemplate` returns null (likely in the case of a new set release with a dungeon run type mode).

Hearthstone Deck Tracker/Hearthstone/DungeonRun.cs
```diff
@@ -22,12 +22,24 @@ public static Deck GetDefaultDeck(string playerClass, CardSet set)
...
 			deck.Name = Helper.ParseDeckNameTemplate(template, deck);
 			return deck;
 		}
 
+		private static string getDeckTemplate(CardSet set)
```
azeier: Method names should be capitalized.

Hearthstone Deck Tracker/Hearthstone/DungeonRun.cs
daltoncwu: Definitely! I'm planning to make `GetDefaultDeck` return null in that case, since the one reference from `DeckManager` already handles that

azeier:
:+1: Looks good

Issue #3830, opened 2018-12-01
Secret tracker erroneously rules out Freezing Trap when the attacking minion dies by Explosive Trap


riQQ:
### Expected Behavior
Secret tracker should not rule out Freezing Trap when the attacking minion dies by Explosive Trap.

### Actual Behavior
Secret tracker erroneously rules out Freezing Trap when the attacking minion dies by Explosive Trap. Explosive Trap was played before Freezing Trap.

### Steps to reproduce behavior
See above

riQQ:
Fixed in [v1.7.15](https://github.com/HearthSim/Hearthstone-Deck-Tracker/releases/tag/v1.7.15)

PR #3833, opened 2018-12-04
Fix a Freezing Trap issue #3830

junsix:

In https://hearthstone.gamepedia.com/Advanced_rulebook#Combat , blizzard write that
"If you play Explosive Trap then Freezing Trap, your enemy's Novice Engineer attacking into it will trigger the Explosive Trap and be reduced to -1 Health. The Freezing Trap is a hindering Secret and thus aborts having no target it considers valid.[30] "


azeier:
Did you confirm this works? I would guess that at the time this code is executed the entities health hasn't changed yet.

junsix:
I confirm that the the entities health is changed at that time.

azeier:
Cool. Can you fix the broken test (I'm gonna guess the test minion has no health, so by adding some), and add another test to specifically test that this happens when the health is < 1?

junsix:
Yes, I'll fix the broken test and add another test to specifically test that this happens when the health is < 1.

azeier:
Great! Thank you!

PR #3842, opened 2018-12-20
Troll set icon

andburn:
Also, noticed the standard icon on deck screenshots is the raven, switched it to the book 🎅


azeier:
Thank you!

Issue #3963, opened 2019-09-17
Ice Barrier is not correctly grayed out when Vaporize triggered


willmiao:
## Bug report
Ice Barrier is not correctly grayed out when a minion tries to attack enemy hero but is killed by Vaporize.

### Expected Behavior

Ice Barrier should be grayed out if it is not triggered in this case.

### Actual Behavior

Ice Barrier is not grayed out

### Steps to reproduce behavior

1. The opponent has two active mage secrets, one of them is Vaporize, the other is not Ice Barrier
2. The player's minion attacks the opponent's hero
3. Vaporize is triggered and the minion died

PR #3964, opened 2019-09-17
Fix secret handling when a minion tries to attack hero but is removed from play

### LINKED ISSUE #3961: Flame Ward incorrectly removed from possible secrets list
Flame Ward secret was incorrectly removed from the list of possible secrets. 

Can't check it properly, but it looks like its is related to opponent having both Flame Ward and Vaporize. As Vaporize have been triggered opponent hero was not properly attacked and thus Flame Ward have not been triggered, however HDT think that this is enough to check if Flame Ward is in play.

Game where it had happened: [https://hsreplay.net/replay/4FgA97N2wqAtvrSEsUYsSE](https://hsreplay.net/replay/4FgA97N2wqAtvrSEsUYsSE)

willmiao:

This fixes #3961 

A minion who tries to attack enemy hero might be removed from play due to:
1. Vaporize or Freezing Trap: tag SHOULDEXITCOMBAT is set to 1
2. Explosive Trap: minion's health less than 1

In these cases, the following secrets won't be triggered:
1. Misdirection(if Vaporize is played first)
2. Sudden Betrayal
3. Flame Ward
4. Freezing Trap(if Vaporize is played first, or if Explosive Trap that kills the minion is triggered first)
5. Vaporize(if Freezing Trap is played first,  or if Explosive Trap that kills the minion is triggered first)

Secret handling is now fixed for these cases.

TODO:
Ice Barrier is not correctly grayed out when I do the tests. See #3963

### DIFF
--- a/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
+++ b/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
@@ -370,7 +370,7 @@ namespace HDTTests.Hearthstone.Secrets
 			_game.SecretsManager.HandleAttack(_playerMinion1, _heroOpponent);
 			VerifySecrets(0, HunterSecrets.All, HunterSecrets.ExplosiveTrap,
 				HunterSecrets.Misdirection, HunterSecrets.WanderingMonster);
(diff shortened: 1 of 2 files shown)


beheh:
@willmiao Why did you close this? Is there any issue or outstanding work?

PR #3966, opened 2019-10-03
Fix HSReplay.net deck short id generation

beheh:
This PR fixes short id generation to mirror https://github.com/HearthSim/hsredshift/pull/169. Previously we used an incorrect comparison method, whereas we want a mostly compatible Unicode Collation Algorithm. For us this means in practice we want a case-insensitive comparison step before the case-sensitive comparison.

Internally refs HSR-462.

### DIFF
--- a/HDTTests/HDTTests.csproj
+++ b/HDTTests/HDTTests.csproj
@@ -116,6 +116,7 @@
     <Compile Include="Hearthstone\WebImportTest.cs" />
     <Compile Include="Hearthstone\DeckTest.cs" />
     <Compile Include="Hearthstone\CardDbTest.cs" />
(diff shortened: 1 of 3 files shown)


beheh:
@azeier Any idea why this PR causes extra tests to run? :thinking: 

Compare
[`Total tests: 450. Passed: 450. Failed: 0. Skipped: 0`](https://ci.appveyor.com/project/HearthSim/hearthstone-deck-tracker/builds/27866788#L3241)
with the previous commit
[`Total tests: 293. Passed: 293. Failed: 0. Skipped: 0`](https://ci.appveyor.com/project/HearthSim/hearthstone-deck-tracker/builds/27606670#L3026)

Am I doing something weird with my imports?

PR #4037, opened 2020-04-20
Fix opponent cards tracking

### LINKED ISSUE #3988: Bombs placed in opponent's deck aren't being counted properly when drawn.
### Expected Behavior
When my opponent draws a bomb, I expect the counter of bombs in their decklist to go down by 1.

### Actual Behavior
When my opponent draws a bomb, the number of bombs in their decklist doesn't change.

Play Bomb generating cards and wait for them to be drawn.

willmiao:

This PR fixes #3988 and other similar problems(e.g., miscount of dream portals shuffled to opponent's deck by Ysera, Unleashed)

### DIFF
--- a/Hearthstone Deck Tracker/LogReader/Handlers/PowerHandler.cs
+++ b/Hearthstone Deck Tracker/LogReader/Handlers/PowerHandler.cs
@@ -179,6 +179,9 @@ namespace Hearthstone_Deck_Tracker.LogReader.Handlers
 					if(string.IsNullOrEmpty(entity.CardId))
 					{
 						entity.CardId = cardId;
+					}
+					if(type == "SHOW_ENTITY")
+					{
 						if(entity.Info.GuessedCardState != GuessedCardState.None)
 							entity.Info.GuessedCardState = GuessedCardState.Revealed;
 					}


azeier:
Can you explain why this shouldn't have the `string.IsNullOrEmpty(entity.CardId)` check? IIRC `SHOW_ENTITY` isn't guaranteed to occur only once.

This seems like a simple change but I'll probably need to spend some time thinking about this before merging it.

willmiao:
Thanks for the reply.

As I understand it:
1. compared to an empty `CardId`, `SHOW_ENTITY`/`HIDE_ENTITY` indicates more precisely whether or not it's ok to show an entity's current info to the player
2. `SHOW_ENTITY` doesn't necessarily have to occur only once. For example, a card is first shown and shuffled into opponent's deck(`SHOW_ENTITY` with `CardId`, `zone`); then its info is held back from the player(`HIDE_ENTITY`) so he won't be able to know whether the card has been drawn or not; at last, when the card is played, cast when drawn or drawn when the hand is full, it is shown to the player again(`SHOW_ENTITY`)

Please let me know if I understand it wrong. Btw, can we make some good use of the tag `REVEALED`?

azeier:
I see what happened. This bug was introduced in https://github.com/HearthSim/Hearthstone-Deck-Tracker/commit/971ebf5414fd22093509848d03dcdf4543471ee7. The `GuessedCardState` behavior was exactly this previous to that commit, as `type` can only be `SHOW_ENTITY` or `CHANGE_ENTITY` here.

Thanks for the fix!

PR #4049, opened 2020-04-29
Implement Madame Lazul card tracking

willmiao:

This PR implements cards tracking for `Madame Lazul` and other similar cards(e.g. `Mind Vision`).

The tag `COPIED_FROM_ENTITY_ID` is used when:
1. discover or copy cards from the opponent's hand, e.g. `Maname Lazul`, `Mind Vision`  <== _This is what we want to keep track of_
2. discover or copy cards from the opponent's deck, e.g. `Cloning Device`, `Thoughtsteal`
3. discover cards from own deck, e.g. `Tracking`, `Sightless Watcher`

Since it doesn't help much knowing which cards are in the opponent's deck, here we only keep track of known cards that are in the opponent's hand. And also, we don't want to handle cards copying initiated by the opponent to avoid info leak.

### DIFF
--- a/Hearthstone Deck Tracker/GameEventHandler.cs
+++ b/Hearthstone Deck Tracker/GameEventHandler.cs
@@ -974,6 +974,11 @@ namespace Hearthstone_Deck_Tracker
 			Core.UpdateOpponentCards();
 		}
 
+		public void HandleCardCopy()
+		{
+			Core.UpdateOpponentCards();
+		}
+
 		public void HandlePlayerJoust(Entity entity, string cardId, int turn)
 		{
 			_game.Player.JoustReveal(entity, turn);
--- a/Hearthstone Deck Tracker/IGameHandler.cs
+++ b/Hearthstone Deck Tracker/IGameHandler.cs
@@ -85,5 +85,6 @@ namespace Hearthstone_Deck_Tracker
 		void HandleChameleosReveal(string cardId);
 		void HandleBeginMulligan();
 		void HandlePlayerMulliganDone();
+		void HandleCardCopy();
 	}
 }
--- a/Hearthstone Deck Tracker/LogReader/Handlers/TagChangeActions.cs
+++ b/Hearthstone Deck Tracker/LogReader/Handlers/TagChangeActions.cs
@@ -75,7 +75,7 @@ namespace Hearthstone_Deck_Tracker.LogReader.Handlers
 			if(!game.Entities.TryGetValue(value, out var targetEntity))
 				return;
 
-			if(string.IsNullOrEmpty(targetEntity.CardId) && targetEntity.IsInHand)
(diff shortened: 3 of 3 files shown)


azeier:
Great PRs! Very much appreciated.

`COPIED_FROM_ENTITY_ID` is also used in Battlegrounds a lot, but this has no impact on that.

> Since it doesn't help much knowing which cards are in the opponent's deck  

Isn't that something HDT does a lot though? The majority of the [card tracking logic in the power handler](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/Hearthstone%20Deck%20Tracker/LogReader/Handlers/PowerHandler.cs#L285) is for cards that get added to decks. Quite possible I'm missing something, but it seems to me that also tracking card copied via `Cloning Device` or `Thoughtsteal` would be more consistent.

willmiao:
> Isn't that something HDT does a lot though? The majority of the card tracking logic in the power handler is for cards that get added to decks. Quite possible I'm missing something, but it seems to me that also tracking card copied via Cloning Device or Thoughtsteal would be more consistent.

True. I implement it that way with no in-hand check first. But after I've "stolen" quite some cards from the opponent in a game I started to forget which of them are in his hand and which are in his deck.
IMO for most of the time, we know what cards are probably in the opponent's deck, what we want to know for sure is whether he has drawn them.

Or maybe is there a better way to implement this, so that we track cards from both hand and deck, and also can tell where are they from?

azeier:
Yeah, maybe we could separate `GuessedCardState.Guessed` into `GuessedCardState.GuessedInHand` and `GuessedCardState.GuessedInDeck` and then distinguish that in the UI with an icon? The only important part here is that we don't dynamically determine where it is, since then you'd likely be able to see it move from deck to hand.

willmiao:
Yes, we should not introduce any info leak.
One possible solution might be we use the new icon for the cards we **definitely** know are in the opponent's hand. If later for some reason(e.g. `Plot Twist`) the cards are shuffled back into the deck, we clear the icon.

azeier:
Yeah something like that. I have a feeling that might be a bit of a bigger project that required a bit more testing, since the way HDT builds the card lists is a bit of a mess.

I'd say we for now include the cards added to the deck as well without any visual separation, since that's more in line with the current HDT behavior and then visual separation of "guessed in hand" vs "guessed in deck" can be added as a separate project and to other effects as well.

Can you modify the PR accordingly?

willmiao:
I deleted the check `targetEntity.IsInHand`, so now we track all known cards either in the hand or in the deck when copy cards.

azeier:
👍 
Is there any reason we might want to check for `IsInHand || IsInDeck`? I can't immediately think of anything, but I'll sleep on it and merge this tomorrow morning!

PR #4127, opened 2020-08-11
Add handling for the korean millhouse hero power

Jonysegal:


HearthWatcher/LogReader/LogFileWatcher.cs
```diff
@@ -24,6 +24,11 @@ public class LogFileWatcher
 		private DateTime _startingPoint;
 		private bool _stop;
 		private Thread _thread;
+		/**When the Millhouse hero power is logged in the korean localization there is a line break in the entity name.
+		 * This causes the parser to get stuck on this broken line and stop reading new power information.
+		 */
```
azeier, line 29:
```suggestion
		/**
		 * When the Millhouse hero power is logged in the korean localization there is a line break in the entity name.
		 * This causes the parser to get stuck on this broken line and stop reading new power information.
		 */
```

PR #4138, opened 2020-08-25
Map transformed aranna cardid to old aranna cardid

Jonysegal:


Hearthstone Deck Tracker/Hearthstone/GameV2.cs
```diff
@@ -31,6 +31,12 @@ public class GameV2 : IGame
 		private Mode _currentMode;
 		private BrawlInfo _brawlInfo;
 		private BattlegroundRatingInfo _battlegroundsRatingInfo;
+		const string UntransformedArannaCardid = HearthDb.CardIds.NonCollectible.Neutral.ArannaStarseekerTavernBrawl1;
+		const string TransformedArannaCardid = HearthDb.CardIds.NonCollectible.Neutral.ArannaStarseeker_ArannaUnleashedTokenTavernBrawl;
+		private Dictionary<string, string> _lastKnownBoardStateLookup = new Dictionary<string, string>()
+		{
+			{ TransformedArannaCardid, UntransformedArannaCardid}
+		};
```
azeier, line 39: Instead of adding more things go GameV2 here, let's make a new `BattlegroundsBoardState` class that contains all this and the `LastKnownBattlegroundsBoardState` dictionary (which can be renamed to something simpler then). 

The class should probably look something like this
```csharp
public void UpdateSnapshot(string opponentHeroCardId); // looks up the key and sets it in the dict

public BoardSnapshot GetSnapshot(string  opponentHeroCardId); // looks up the key and returns the result from the dict
```

This will make it completely opaque to the implementation (GameV2) that there is any card id manipulation going on, and nothing but that class will need to care how the snapshots are keyed.

Hearthstone Deck Tracker/Hearthstone/GameV2.cs
```diff
@@ -301,7 +307,9 @@ public void SnapshotBattlegroundsBoardState()
 				.Select(x => x.Clone())
 				.ToArray();
 			Log.Info($"Snapshotting board state for {opponentHero.Card.Name} with {entities.Length} entities");
-			LastKnownBattlegroundsBoardState[opponentHero.CardId] = new BoardSnapshot(entities, GetTurnNumber());
+			LastKnownBattlegroundsBoardState[GetCorrectLastKnownBoardStateCardId(opponentHero.CardId)] = new BoardSnapshot(entities, GetTurnNumber());
 		}
+
+		internal string GetCorrectLastKnownBoardStateCardId(string cardId) => _lastKnownBoardStateLookup.TryGetValue(cardId, out var mapped) ? mapped : cardId;
```
azeier, line 313: `Dictionary.TryGetValue` will throw an `ArgumentNullException` when cardId is null, so you should add a check for that.

Hearthstone Deck Tracker/Hearthstone/BattlegroundsBoardState.cs
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		public BoardSnapshot GetSnapshot(string opponentHeroCardId) => LastKnownBoardState.TryGetValue(opponentHeroCardId, out var state) ? state : null;
+
+		public void Reset()
+		{
+			LastKnownBoardState.Clear();
+		}
+
+		internal string GetCorrectLastKnownBoardStateCardId(string cardId) => cardId != null && _lastKnownBoardStateLookup.TryGetValue(cardId, out var mapped) ? mapped : cardId;
```
azeier, line 51: This can be made private and we can simplify the name a bit to e.g. `GetBoardStateKey`.

Hearthstone Deck Tracker/Hearthstone/GameV2.cs
```diff
@@ -31,13 +31,16 @@ public class GameV2 : IGame
 		private Mode _currentMode;
 		private BrawlInfo _brawlInfo;
 		private BattlegroundRatingInfo _battlegroundsRatingInfo;
+		private BattlegroundsBoardState _battlegroundsBoardState;
+
```
azeier, line 35:
```suggestion
```

Hearthstone Deck Tracker/Hearthstone/IGame.cs
```diff
@@ -1,6 +1,7 @@
 #region
 
 using System.Collections.Generic;
+using System.Collections.Specialized;
```
azeier, line 4:
```suggestion
```

Hearthstone Deck Tracker/Hearthstone/GameV2.cs
```diff
@@ -289,19 +291,8 @@ public int GetTurnNumber()
...
-				.Where(x => x.IsMinion && x.IsInZone(Zone.PLAY) && x.IsControlledBy(Opponent.Id))
-				.Select(x => x.Clone())
-				.ToArray();
-			Log.Info($"Snapshotting board state for {opponentHero.Card.Name} with {entities.Length} entities");
-			LastKnownBattlegroundsBoardState[opponentHero.CardId] = new BoardSnapshot(entities, GetTurnNumber());
-		}
-	}
+		public void SnapshotBattlegroundsBoardState() => _battlegroundsBoardState.UpdateSnapshot();
+
+		public BoardSnapshot GetBattlegroundsBoardStateFor(string cardId) => _battlegroundsBoardState.GetSnapshot(cardId);
```
azeier, line 296: I think we might as well get rid of these at this point, make `BattlegroundsBoardState` internal and access that directly. That will also make the changes to MockGame unnecessary..

PR #4208, opened 2021-01-08
Add opponent dead for turns

Jonysegal:


Hearthstone Deck Tracker/Windows/OverlayWindow.xaml.cs
```diff
@@ -370,7 +371,20 @@ internal async Task ExperienceChangedAsync(int experience, int experienceNeeded,
...
+		{
+			var index = _leaderboardDeadForText.Count - 1;
+			foreach(var text in _leaderboardDeadForText)
+				text.Text = "";
+			foreach(var text in _leaderboardDeadForTurnText)
+				text.Text = "";
+			foreach(var turn in turns)
+			{
+				_leaderboardDeadForText[index].Text = $"{turn}";
+				_leaderboardDeadForTurnText[index].Text = "turns";
```
beheh, line 385: Can we make this translateable?

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+			{
+				_uniqueDeadHeroes.Clear();
+				_deadTracker.Clear();
+				Core.Overlay.nextOpponentLeaderboardPosition = null;
+			}
+			for(int i =0; i < _deadTracker.Count; i++)
+			{
+				_deadTracker[i]++;
+			}
+			var deadHeroes = game.Entities.Select(x=>x.Value).Where(x => x.IsHero && x.Health <= 0);
```
azeier, line 26:
```suggestion
			var deadHeroes = game.Entities.Values.Where(x => x.IsHero && x.Health <= 0);
```

Hearthstone Deck Tracker/Windows/OverlayWindow.xaml.cs
azeier, line 385: Needs localization, and a singular. Something like `turns == 1 ? GetLoc(FOO_TURN) : GetLoc(FOO_TURNS)`

azeier, line 385: Using an index that is not based on the size of the array is a bit dangerous. You should verify that the index is in range of both of the arrays.

Hearthstone Deck Tracker/Windows/OverlayWindow.xaml.cs
```diff
@@ -370,7 +371,20 @@ internal async Task ExperienceChangedAsync(int experience, int experienceNeeded,
 			}
 			if(_game.CurrentMode != Enums.Hearthstone.Mode.HUB)
 				HideExperienceCounter();
-
+		}
```
azeier, line 374:
```suggestion
		}

```

Hearthstone Deck Tracker/Windows/OverlayWindow.MouseOverDetection.cs
```diff
@@ -85,15 +87,35 @@ private void UpdateMouseOverDetectionRegions(List<Entity> oppBoard, List<Entity>
...
+				Canvas.SetTop(_leaderboardIcons[i], LeaderboardTop + BattlegroundsTileHeight * i);
 				Canvas.SetLeft(_leaderboardIcons[i], Helper.GetScaledXPos(0.001 * (_leaderboardIcons.Count - i - 1), (int)Width, ScreenRatio));
 			}
+
+			PositionDeadForText();
+		}
+
+		internal void PositionDeadForText()
+		{
+			var leftAdjust = .00075;
```
azeier, line 104: Can be a `const`. A short comment explaning what this does would be good.

Hearthstone Deck Tracker/Windows/OverlayWindow.MouseOverDetection.cs
```diff
@@ -85,15 +87,35 @@ private void UpdateMouseOverDetectionRegions(List<Entity> oppBoard, List<Entity>
...
+			{
+				Canvas.SetTop(_leaderboardDeadForText[i], LeaderboardTop + BattlegroundsTileHeight * i);
+				Canvas.SetLeft(_leaderboardDeadForText[i], Helper.GetScaledXPos(leftAdjust * (_leaderboardDeadForText.Count - i - 1), (int)Width, ScreenRatio));
+				Canvas.SetTop(_leaderboardDeadForTurnText[i], LeaderboardTop + BattlegroundsTileHeight * i);
+				Canvas.SetLeft(_leaderboardDeadForTurnText[i], Helper.GetScaledXPos(leftAdjust * (_leaderboardDeadForTurnText.Count - i - 1), (int)Width, ScreenRatio));
+			}
+
+			if(nextOpponentLeaderboardPosition != null)
+			{
+				Canvas.SetLeft(_leaderboardDeadForText[(nextOpponentLeaderboardPosition ?? 0)-1], Helper.GetScaledXPos(leftAdjust * (_leaderboardDeadForText.Count - (nextOpponentLeaderboardPosition ?? 0) - 1 - 1) + .025, (int)Width, ScreenRatio));
```
azeier, line 116: You should and can get rid of the null coalescing operator here. For one because you check for null above and also since if it was null you would throw an `IndexOutOfRangeException` with `_leaderboardDeadForText[-1]`

Hearthstone Deck Tracker/Windows/OverlayWindow.MouseOverDetection.cs
```diff
@@ -85,15 +87,35 @@ private void UpdateMouseOverDetectionRegions(List<Entity> oppBoard, List<Entity>
...
+				Canvas.SetLeft(_leaderboardDeadForTurnText[(nextOpponentLeaderboardPosition ?? 0)-1], Helper.GetScaledXPos(leftAdjust * (_leaderboardDeadForTurnText.Count - (nextOpponentLeaderboardPosition ?? 0) - 1 - 1) + .025, (int)Width, ScreenRatio));
```
azeier, line 117: `leftAdjust * (_leaderboardDeadForTurnText.Count - (nextOpponentLeaderboardPosition ?? 0) - 1 - 1) + .025`
I find it a bit hard to grasp what is happening here and why. Again, you can drop the null coalescing operator, maybe try to pull some parts of it out into well named variables and add some comments if nothing else. What's `0.025`? Why `-1 -1`?

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+				if(currentPlayer != null && currentPlayer.HasTag(GameTag.NEXT_OPPONENT_PLAYER_ID))
+				{
+					var nextOpponent = gameEntites.FirstOrDefault(x => x.GetTag(GameTag.PLAYER_ID) == currentPlayer.GetTag(GameTag.NEXT_OPPONENT_PLAYER_ID));
+					if(nextOpponent != null)
+					{
+						var leaderboardPlace = nextOpponent.GetTag(GameTag.PLAYER_LEADERBOARD_PLACE);
+						if(leaderboardPlace >= 0 && leaderboardPlace < 8)
+						{
+							Core.Overlay.nextOpponentLeaderboardPosition = leaderboardPlace;
+							Core.Overlay.PositionDeadForText();
```
azeier, line 51: Instead of updating a field on another class, that is used for a subsequent call, calling `Core.Overlay.PositionDeadForText(leaderboardPlace)` would be a lot cleaner.

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+				{
+					_deadTracker.Add(0);
+					_uniqueDeadHeroes.Add(id);
+				}
+			}
+			_deadTracker.Sort((x, y) => y.CompareTo(x));
+			Core.Overlay.UpdateOpponentDeadForTurns(_deadTracker);
+			var gameEntites = game.Entities.Values;
+			var currentPlayer = gameEntites.FirstOrDefault(x => x.IsCurrentPlayer);
+			for(int i = 0; i < 5; i++)
```
azeier, line 40: A brief comment explaning why this loop is here would be good.

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
azeier, line 51: I believe this is the success case? You can `break` the for loop here to avoid more iterations.

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+		private static List<int> _deadTracker = new List<int>();
+		const string KelThuzadCardId = "KelThuzad";
+
+		public static async void ShoppingStarted(GameV2 game)
+		{
+			if(game.GetTurnNumber() <= 1)
+			{
+				_uniqueDeadHeroes.Clear();
+				_deadTracker.Clear();
+				Core.Overlay.nextOpponentLeaderboardPosition = null;
```
azeier, line 22: Related to the comment below, calling something like `Core.Overlay.ClearPositionDeadForText()` would be better than setting a field to null.

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
```diff
@@ -0,0 +1,60 @@
...
+			}
+			for(int i =0; i < _deadTracker.Count; i++)
```
azeier, line 24:
```suggestion
			for(int i = 0; i < _deadTracker.Count; i++)
```

azeier (changes requested):

Hearthstone Deck Tracker/OpponentDeadForTracker.cs
Jonysegal: We want to scan to the end of the loop because at the start of shopping that will be hit by the old opponent, so we want to wait until the very end to get the next opponent. Maybe this should just wait 2500 at the beginning and not loop; I added the loop thinking it would make it hit as soon as possible but might not be worth the trouble

azeier: I see, makes sense. Adding that to the comment would be good as well.

azeier (approved):

PR #4209, opened 2021-01-08
Cthun data leak

Jonysegal:


azeier (approved):

lacerbeams:
Now I cannot see my own C'thun cards in my hand

PR #4213, opened 2021-01-19
Update card tooltips

Jonysegal:


Hearthstone Deck Tracker/Utility/RemoteConfig.cs
```diff
@@ -135,6 +138,12 @@ internal class BobsBuddyData
 				[JsonProperty("log_lines_kept")]
 				public int LogLinesKept { get; set; }
 			}
+
+			internal class UpdateData
+			{
+				[JsonProperty("image_update_hearthstone_version")]
+				public string Version { get; set; }
```
azeier, line 145: We should just make this a `int` value that we increment as needed. We a) don't want to maintain this version with every hearthstone patch and b) might want to update it multiple times per hearthstone patch is something goes wrong with the images for some reason.

Hearthstone Deck Tracker/Controls/Card.xaml
```diff
@@ -4,12 +4,15 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:controls="clr-namespace:Hearthstone_Deck_Tracker.Controls"
+             xmlns:hearthstone="clr-namespace:Hearthstone_Deck_Tracker.Hearthstone"
```
azeier, line 7: This appears to be unused?

Hearthstone Deck Tracker/Controls/Card.xaml.cs
```diff
@@ -19,5 +19,21 @@ public bool HasTooltip
...
+		private void Rectangle_ToolTipOpening(object sender, System.Windows.Controls.ToolTipEventArgs e)
+		{
+			TooltipCardImage.SetCardId(CardId());
+		}
+
+		public string CardId()
+		{
+			var card = DataContext as Hearthstone.Card;
+			return card?.Id;
+		}
```
azeier, line 32: Can be simplified as
```suggestion
		public string CardId => (DataContext as Hearthstone.Card)?.CardId;
```

Hearthstone Deck Tracker/Controls/CardImage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+		{
+			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+		}
+
+		public CardImage()
+		{
+			InitializeComponent();
+		}
+
+		public async void SetCardId(string cardId)
```
azeier, line 72: For readability purposes I would suggest renaming `cardId` to `newCardId`. `cardId` and `CardId` are difficult to distinguish at a glance.

Hearthstone Deck Tracker/Controls/CardImage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+
+		public event PropertyChangedEventHandler PropertyChanged;
+
+		[NotifyPropertyChangedInvocator]
+		protected virtual void OnPropertyChanged([CallerMemberName] string propertyName = null)
+		{
+			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+		}
```
azeier, line 65: Since I remembered yesterday that it exists, let's just delete this and the `INotifyProperty` inheritance and replace it with `ViewModel`.

Hearthstone Deck Tracker/Controls/CardImage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+			if(string.IsNullOrEmpty(cardId))
+			{
+				CardImagePath = null;
+				return;
+			}
+			if(!AssetDownloaders.cardImageDownloader.HasAsset(CardId))
+			{
+				await AssetDownloaders.cardImageDownloader.DownloadAsset(CardId);
+			}
+			CardImagePath = AssetDownloaders.cardImageDownloader.StoragePathFor(CardId);
```
azeier, line 86: There is something odd that can happen here when `SetCardId` is called a second time before the `await DownloadAsset()` completes but finishes faster than the first call. `CardId` will be set to a new value and `StoragePathFor()` will be invoked with the id from the *second* call.

While this is actually the behaviour we want, it happens in a way that isn't very obvious. This method should instead use `newCardId` throughout and have a check after the download so see whether we should still update the CardImagePath (i.e. exit if `newCardId` is now different from `CardId`).

Hearthstone Deck Tracker/Core.cs
```diff
@@ -169,6 +173,38 @@ public static async void Initialize()
...
+				}
+			}
+			catch(Exception ex)
+			{
+				Log.Error($"Could not check for updated HearthDB version: {ex}");
+			}
+			try
+			{
+				var hearthDbVersion = HearthDb.Info.HearthDbVersion.ToString();
+				if(RemoteConfig.Instance?.Data?.UpdateInfo?.Version != Config.Instance.RemoteHearthstoneVersion)
```
azeier, line 195: Remote will not be loaded yet at this point. `RemoteConfig.Instance.Load` is called near the end of `Core.Initialize`. What you can do is move this second try-catch into it's own method and hook into the `RemoteConfig.Instance.Loaded` event *before* the `RemoteConfig.Instance.Load` call with this method. (`RemoteConfig.Instance.Loaded += YourNewCheck`)

Hearthstone Deck Tracker/FlyoutControls/Options/Tracker/TrackerGeneral.xaml
```diff
@@ -4,12 +4,15 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
              xmlns:options="clr-namespace:Hearthstone_Deck_Tracker.FlyoutControls.Options"
-             xmlns:lex="http://wpflocalizeextension.codeplex.com"
+             xmlns:lex="http://wpflocalizeextension.codeplex.com" xmlns:converters="clr-namespace:Hearthstone_Deck_Tracker.Controls.Stats.Converters"
```
azeier, line 7:
```suggestion
             xmlns:lex="http://wpflocalizeextension.codeplex.com"
             xmlns:converters="clr-namespace:Hearthstone_Deck_Tracker.Controls.Stats.Converters"
```

Hearthstone Deck Tracker/Windows/OverlayWindow.Tooltips.cs
```diff
@@ -69,33 +29,33 @@ private void UpdateCardTooltip()
...
+			
+			ToolTipCardBlock.CreatedByVisibility = Collapsed;
 			if(!Config.Instance.HideOpponentCardMarks && cardMark != null)
 			{
 				var index = _cardMarks.IndexOf(cardMark.Label);
 				var card = _game.Opponent.Hand.FirstOrDefault(x => x.GetTag(GameTag.ZONE_POSITION) == index + 1 && x.HasCardId && !x.Info.Hidden)?.Card;
 				var creatorCard = _cardMarks[index].SourceCard;
-				if(card != null || creatorCard != null)
+				if(card == null && creatorCard != null)
 				{
```
azeier, line 45: I think this check needs to stay the way it was? There is two "types" of tooltips we want to show here. 1) For cards where we know the creator we want to show the creator card and the "Created by {creator}" label. 2) For cards where we know the cardId, e.g. because the card was returned to hand or we were able to predict it, we want to show that card but *hide* the created by label.

Hearthstone Deck Tracker/Config.cs
```diff
@@ -732,9 +738,11 @@ public class Config
 		[DefaultValue("enUS")]
 		public string SelectedLanguage = "enUS";
 
-		[XmlArray(ElementName = "AlternativeLanguages")]
-		[XmlArrayItem(ElementName = "Language")]
-		public List<string> AlternativeLanguages = new List<string>();
+		[DefaultValue(false)]
+		public bool LargeCardImages = false;
```
azeier, line 742: Let's turn this into a `double` with `DefaultValue(1)` and add a dropdown or slider (depending on what is easier) to the settings instead of the checkbox with `1x, 1.25x, 1.5x, 1.75x, 2x`.

With a slider I believe you can set `min=1, max=2, step=0.25`

azeier (changes requested):

Jonysegal:
Need to add localization

Hearthstone Deck Tracker/Core.cs
```diff
@@ -169,6 +175,42 @@ public static async void Initialize()
...
+					Config.Save();
+				}
+			}
+			catch(Exception ex)
+			{
+				Log.Error($"Could not check for updated HearthDB version: {ex}");
+			}
+		}
+
+		private static void CheckForCardImageUpdate()
```
azeier, line 196: You can simplify all of this a little bit by doing this:
```diff
index c1cc788c..ec3343aa 100644
--- a/Hearthstone Deck Tracker/Core.cs	
+++ b/Hearthstone Deck Tracker/Core.cs	
@@ -152,7 +152,7 @@ namespace Hearthstone_Deck_Tracker
 			}
 			LogWatcherManger.Start(Game).Forget();
 
-			RemoteConfig.Instance.Loaded += (ConfigData data) => CheckForCardImageUpdate();
+			RemoteConfig.Instance.Loaded += CheckForCardImageUpdate;
 			RemoteConfig.Instance.Load();
 			HotKeyManager.Load();
 
@@ -193,15 +193,16 @@ namespace Hearthstone_Deck_Tracker
 			}
 		}
 
-		private static void CheckForCardImageUpdate()
+		private static void CheckForCardImageUpdate(ConfigData rConfig)
 		{
 			try
 			{
 				var hearthDbVersion = HearthDb.Info.HearthDbVersion.ToString();
-				if(RemoteConfig.Instance?.Data?.UpdateInfo?.Version != Config.Instance.RemoteHearthstoneVersion)
+				var remoteVersion = rConfig?.UpdateInfo?.Version;
+				if(remoteVersion.HasValue && remoteVersion > Config.Instance.RemoteHearthstoneVersion)
 				{
 					AssetDownloaders.cardImageDownloader.ClearStorage();
-					Config.Instance.RemoteHearthstoneVersion = RemoteConfig.Instance?.Data?.UpdateInfo?.Version ?? 0;
+					Config.Instance.RemoteHearthstoneVersion = remoteVersion.Value;
 					Config.Save();
 				}
 			}
```

- Loaded is called with the config data already so you don't need to manually look it up again
- We should explicitly check that ` remoteVersion > localVersion` rather than `remoteVersion != localVersion`

Hearthstone Deck Tracker/Windows/OverlayWindow.Tooltips.cs
```diff
@@ -69,33 +29,33 @@ private void UpdateCardTooltip()
...
+					ToolTipCardBlock.CreatedByVisibility = Visible;
+					ToolTipCardBlock.CreatedByText = $"Created By {creatorCard.Name}";
+					ToolTipCardBlock.SetCardIdFromCard(creatorCard);
 					var offset = _cardMarks[index].ActualHeight * 1.25;
 					var topOffset = Canvas.GetTop(_cardMarks[index]) + offset;
 					var leftOffset = Canvas.GetLeft(_cardMarks[index]) + offset;
-					Canvas.SetTop(ToolTipCard, topOffset);
-					Canvas.SetLeft(ToolTipCard, leftOffset);
-					ToolTipCard.Visibility = Config.Instance.OverlayCardMarkToolTips ? Visible : Hidden;
-					ToolTipCard.CreatedByVisibility = card != null ? Collapsed : Visible;
```
azeier, line 93: I believe we still need this?

azeier (approved):

PR #4279, opened 2021-04-27
Support multiclass cards with StringImporter

### LINKED ISSUE #4273: StringImporter not importing some decks that contain multiclass cards
### Expected Behavior
All decks should be imported correctly, and if the player class is not clear, let the player choose.

### Actual Behavior
The StringImported.Import function is not importing some decks that contain multiclass cards. Currently, all cards are tied to a single class, even if they can be used by two (like the cards from Scholomance).

For example, Lightning Bloom is internally tagged as a DRUID card, but it can be added to Shaman and Druid decks. When using the StringImporter, it detects both Shaman and Druid classes and returns a null deck, discarding it.

This can be tested using the TestNeutralOnly Test:
1. Open ParseCardStringTest.cs (in HDTTests > Hearthstone).
2. Copy the following string from a Shaman deck:
```Lightning Bloom x 2\nDevolving Missiles\nLightning Bolt x 2\nNovice Zapper x 2\nPrimordial Studies x 2\nBloodmage Thalnos\nCagematch Custodian x 2\nDiligent Notetaker x 2\nLandslide x 2\nRockbiter Weapon x 2\nInstructor Fireheart\nSerpentshrine Portal x 2\nStormstrike x 2\nDunk Tank x 2\nTorrent x 2\nDoomhammer x 2\nInara Stormcrash\n```
3. Replace the string "Ragnaros the firelord" with the copied above.
4. Run the test and notice that it fails as the returned deck is null as a result of detecting two different player classes.

### Log/Screenshots
Warning|StringImporter.Import >> Could not identify a class for this deck. Found class cards for 2 classes.

sgkoishi:
Fix #4273

### DIFF
--- a/Hearthstone Deck Tracker/Importing/StringImporter.cs
+++ b/Hearthstone Deck Tracker/Importing/StringImporter.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Text.RegularExpressions;
+using HearthDb.Enums;
 using Hearthstone_Deck_Tracker.Hearthstone;
 using Hearthstone_Deck_Tracker.Utility.Logging;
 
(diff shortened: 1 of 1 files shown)


azeier:
Thanks!

PR #4285, opened 2021-05-11
Add Octobot Reduction Symbol

tim-inzitari:
- [y ] I have read the [Contributing Guidelines](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/CONTRIBUTING.md#contributing)
- [ y] I have already signed the CLA

### DIFF
--- a/Hearthstone Deck Tracker/GameEventHandler.cs
+++ b/Hearthstone Deck Tracker/GameEventHandler.cs
@@ -286,6 +286,12 @@ namespace Hearthstone_Deck_Tracker
 			}
 		}
 
+		public void HandleCostReduction(int amount)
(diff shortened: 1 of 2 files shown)


Jonysegal (changes requested):
I think we should avoid this level of detail in the Power Handler, so basically put that function where the hardcoded cost reduction is. Then also use the HandleCostReduction where HandleThaurissanCostReduction.

Also HandleCostReduction is a bit too general of a name for what the function is doing. With how these Handle functions work it should take an Entity and a value as arguments and apply the modifications to the entity based on the value.

beheh:
Replaced by #4288.

PR #4296, opened 2021-06-07
Added Ysera the Dreamer Dream card tracking

reesercollins:

Adds tracking for the dream cards which are given from Ysera the Dreamer's Battlecry.


Jonysegal:
Sorry for the late response; this will not work because this system adds CardIds for all cards generated under the current block. What that means is that HDT will always think the dream card Ysera generates is the first card you have added to known card ids, in this case Nightmare. The default HDT handling where it is just ascribed to Ysera is better here.

PR #4298, opened 2021-06-10
Fix Reckoning Secret Being Incorrectly Excluded

### LINKED ISSUE #4290: Secret Tracker incorrectly tracks Paladin's Reckoning if the attacking minion died
**Bug report**
Reckoning incorrectly becomes greyed out if a minion with 3+ attack dies to another minion after its attack. Despite what the tracker shows in this instance, Reckoning is still a possible secret outcome.

 **Expected Behavior**
Reckoning stays active in the tracker when the minion with more than three attack dies to another minion.

**Actual Behavior**
Reckoning becomes greyed out when the minion with more than three attack dies to another minion.

 **Steps to reproduce behavior**
Try killing off your 3+ attack minion to an enemy minion (thus not triggering the secret) and the secret will be greyed out on the tracker.

reesercollins:

Fixes #4290

### DIFF
--- a/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
+++ b/HDTTests/Hearthstone/Secrets/SecretEventHandlerTest.cs
@@ -298,6 +298,8 @@ namespace HDTTests.Hearthstone.Secrets
 		public void SingleSecret_OpponentMinionDealtThreeDamage_TriggersReckoning()
 		{
 			SetPlayerAsCurrentPlayer();
+			// Prevent attacker from dying
+			_playerMinion1.SetTag(GameTag.HEALTH, 5);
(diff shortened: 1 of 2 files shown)


Jonysegal:
Thanks for the pr! Coincidentally I was working on a fix for the same problem, but looking at your approach made me think of some more nuance that this fix needs (eg poisonous, divine shield + poisonous, when certain tags are logged), so I'll work on those and push a fix soon.

Jonysegal:
After implementing and testing the divine shield and poisonous fixes I tested it and realized reckoning doesn't actually care about poisonous, so that didn't end up mattering. Regardless, fix with a task delay to allow the relevant power lines to appear is here
https://github.com/HearthSim/Hearthstone-Deck-Tracker/pull/4299

PR #4506, opened 2023-10-25
fix: shuffle Plagues back to deck when affected by Plight Of The Dead

KelvinCS:

### DIFF
--- a/Hearthstone Deck Tracker/LogReader/Handlers/PowerHandler.cs
+++ b/Hearthstone Deck Tracker/LogReader/Handlers/PowerHandler.cs
@@ -535,6 +535,13 @@ namespace Hearthstone_Deck_Tracker.LogReader.Handlers
 								AddKnownCardId(gameState, NonCollectible.Neutral.RivendareWarrider_KorthazzDeathriderToken); 
 								AddKnownCardId(gameState, NonCollectible.Neutral.RivendareWarrider_ZeliekConquestriderToken); 
 								break;
+							case NonCollectible.Deathknight.Helya_PlightOfTheDeadEnchantment:
+								if (gameState.LastPlagueDrawn != null)
+								{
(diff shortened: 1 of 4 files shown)


Hearthstone Deck Tracker/LogReader/Handlers/PowerHandler.cs
```diff
@@ -535,6 +535,13 @@ public void Handle(string logLine, IHsGameState gameState, IGame game)
 								AddKnownCardId(gameState, NonCollectible.Neutral.RivendareWarrider_KorthazzDeathriderToken); 
 								AddKnownCardId(gameState, NonCollectible.Neutral.RivendareWarrider_ZeliekConquestriderToken); 
 								break;
+							case NonCollectible.Deathknight.Helya_PlightOfTheDeadEnchantment:
+								if (gameState.LastPlagueDrawn != null)
+								{
+									AddKnownCardId(gameState, gameState.LastPlagueDrawn);
```
edipo2s, line 541: I think Plagues should be re-added to the deck after drawn only if Helya was played this game because of their ability: `Plagues they draw this game are unending.` You can track if a Helya was played the same way you are tracking the last drawn plague, but it's worth checking if there's some game state like a `tag` value on the plague instance that already indicates that

KelvinCS: We are already verifying that on line 538. We look for Helya's enchantment that brings back the Plague into the deck. If this enchantment is activated, we get the last plague drawn.

Is it possible to activate `Helya_PlightOfTheDeadEnchantment` before playing Helya?

edipo2s: Oh, I see, the trigger comes from his enchantment, all good then

azeier (approved):

PR #4507, opened 2023-10-26
feat: show minions top bar before and during hero picking

KelvinCS:

### DIFF
--- a/Hearthstone Deck Tracker/GameEventHandler.cs
+++ b/Hearthstone Deck Tracker/GameEventHandler.cs
@@ -521,7 +521,7 @@ namespace Hearthstone_Deck_Tracker
 
 			if(_game.IsBattlegroundsMatch && _game.CurrentGameMode == GameMode.Spectator)
 			{
-				Core.Overlay.ShowBgsTopBar();
+				Core.Overlay.ShowBgsTopBarAndBobsBuddyPanel();
 				Core.Overlay.ShowBattlegroundsSession(true);
 			}
 			if(_game.IsFriendlyMatch)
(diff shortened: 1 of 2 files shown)


Hearthstone Deck Tracker/GameEventHandler.cs
```diff
@@ -1072,6 +1072,7 @@ private async void HandleConstructedStart()
 
 		private async void HandleBattlegroundsStart()
 		{
+			Core.Overlay.ShowBgsTopBar();
```
azeier, line 1075: We should wait for the game to be faded in before we show the UI. Below is a 3 second delay; this should probably go into the else case after that.

azeier (changes requested):
There is a small bug here when queuing into multiple games in a row:
- In the first game everything works as expected, minion list is visible during hero picking, BobsBuddy panel is not. Play one combat to have the BobsBuddy panel fully appear, concede.
- In the second game the BobsBuddy panel is now also visible during hero picking.

There is probably some behavior missing to hide the BobsBuddy panel specifically that was not necessary previously, since the entire top section was only visible after hero picking.

KelvinCS:
> There is a small bug here when queuing into multiple games in a row:
> 
> * In the first game everything works as expected, minion list is visible during hero picking, BobsBuddy panel is not. Play one combat to have the BobsBuddy panel fully appear, concede.
> * In the second game the BobsBuddy panel is now also visible during hero picking.
> 
> There is probably some behavior missing to hide the BobsBuddy panel specifically that was not necessary previously, since the entire top section was only visible after hero picking.

Fixed by other PR

PR #4735, opened 2026-06-26
Fix formatted flavor text with XML special characters

### LINKED ISSUE #4734: Flavor text can disappear when it contains XML special characters
## Description

Flavor text shown in the overlay can disappear when the text contains XML special characters such as `&`, `<`, or `>`.

This is probably not very common, but it can happen with card text/localization data. The chance of these characters appearing in card flavor text may be low, but with Blizzard content, who knows :)

The issue is in the formatted flavor text renderer. It wraps the text in XML and parses it with `XElement.Parse`. Formatting tags like `<b>` and `<i>` are expected to work, but plain text XML characters are not escaped first. For example, if the flavor text contains `A & B`, XML parsing fails and the renderer returns no inlines, so the text is not displayed.

## User impact

A user may see an empty flavor text tooltip for specific cards/localizations instead of the actual flavor text.

Example scenario:

1. Flavor text display is enabled.
2. The user hovers a card whose flavor text contains `&`, `<`, or `>`.
3. The overlay tries to render the formatted flavor text.
4. XML parsing fails because the plain text character was not escaped.
5. The tooltip content is empty instead of showing the flavor text.

This does not appear to crash the application, but it causes the affected flavor text to silently disappear.

## Reproduction

I added a regression test for `AttachedFormattedString.SetFormattedText`:

```csharp
AttachedFormattedString.SetFormattedText(textBlock, "A & B");
```

Before the fix, the rendered text is empty instead of `A & B`.

## Root cause

`AttachedFormattedString.GetInlines` wraps the input string in a temporary XML element and parses it:

```csharp
XElement.Parse("<w>" + ... + "</w>")
```

The code preserves supported formatting tags, but it does not escape plain text XML characters before parsing.

## Expected behavior

Plain text XML special characters should be displayed normally, while supported formatting tags such as `<b>` and `<i>` should continue to apply formatting.

## Validation

Target branch/commit:

- `origin/master`
- `9770f02`

Validation performed:

```text
MSBuild.exe .\HDTTests\HDTTests.csproj /p:Configuration=Debug /p:Platform=x86 /m /v:minimal
vstest.console.exe .\HDTTests\bin\x86\Debug\HDTTests.dll /Tests:FormattedText_PreservesAmpersand,FormattedText_PreservesBoldTag /Platform:x86
vstest.console.exe .\HDTTests\bin\x86\Debug\HDTTests.dll /Platform:x86
```

Results:

- Regression test fails before the fix: expected `A & B`, actual empty string.
- Targeted tests after the fix: 2/2 passed.
- Full test suite after the fix: 210/210 passed.

numbereleven-a:
## Summary

Fix formatted flavor text rendering when the text contains XML special characters such as `&`, `<`, or `>`.

The renderer wraps flavor text in a temporary XML element and parses it with `XElement.Parse`. Supported formatting tags like `<b>` and `<i>` should still work, but plain text XML characters need to be escaped first. Without that, a flavor text value like `A & B` fails XML parsing and renders no text.

Fixes #4734.

## User impact

Users can see an empty flavor text tooltip for specific cards/localizations instead of the actual flavor text. This is probably not very common, but it can happen with card text/localization data.

## Changes

- Escape plain text XML characters before parsing formatted flavor text.
- Preserve supported formatting tags: `<b>`, `</b>`, `<i>`, `</i>`.
- Add regression coverage for ampersands and existing bold formatting behavior.

## Validation

Validated against `origin/master` at `9770f02`.

```text
MSBuild.exe .\HDTTests\HDTTests.csproj /p:Configuration=Debug /p:Platform=x86 /m /v:minimal
vstest.console.exe .\HDTTests\bin\x86\Debug\HDTTests.dll /Tests:FormattedText_PreservesAmpersand,FormattedText_PreservesBoldTag /Platform:x86
vstest.console.exe .\HDTTests\bin\x86\Debug\HDTTests.dll /Platform:x86
```

Results:

- Regression test fails before the fix: expected `A & B`, actual empty string.
- Targeted tests after the fix: 2/2 passed.
- Full test suite after the fix: 210/210 passed.

### DIFF
--- a/HDTTests/Utility/AttachedFormattedStringTests.cs
+++ b/HDTTests/Utility/AttachedFormattedStringTests.cs
@@ -1,0 +1,38 @@
+using System.Linq;
+using System.Windows;
+using System.Windows.Controls;
+using System.Windows.Documents;
+using Hearthstone_Deck_Tracker.Utility;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace HDTTests.Utility
+{
+	[TestClass]
+	public class AttachedFormattedStringTests
+	{
+		[TestMethod]
+		public void FormattedText_PreservesAmpersand()
+		{
+			var textBlock = new TextBlock();
+
+			AttachedFormattedString.SetFormattedText(textBlock, "A & B");
+
+			var text = string.Concat(textBlock.Inlines.OfType<Run>().Select(x => x.Text));
+			Assert.AreEqual("A & B", text);
+		}
+
+		[TestMethod]
+		public void FormattedText_PreservesBoldTag()
+		{
+			var textBlock = new TextBlock();
+
+			AttachedFormattedString.SetFormattedText(textBlock, "A <b>B</b>");
+
+			var runs = textBlock.Inlines.OfType<Run>().ToArray();
+			Assert.AreEqual(2, runs.Length);
+			Assert.AreEqual("A ", runs[0].Text);
+			Assert.AreEqual("B", runs[1].Text);
+			Assert.AreEqual(FontWeights.Bold, runs[1].FontWeight);
+		}
+	}
+}
--- a/Hearthstone Deck Tracker/Utility/AttachedFormattedString.cs
+++ b/Hearthstone Deck Tracker/Utility/AttachedFormattedString.cs
@@ -38,7 +38,7 @@ namespace Hearthstone_Deck_Tracker.Utility
 			try
 			{
 				//<w> is a wrapper so it gets read as an XElement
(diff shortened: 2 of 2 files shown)


beheh:
Is there any actual case of this? This feels a little overlay defensive at this time, introducing code complexity for a change that may not come for a long time (or ever).

This fix feels a little too speculative. Did this come from a genuine case, or e.g. from an LLM doing speculative research for issues in this code base?

numbereleven-a:
Of course there are no examples, there are no such playing cards yet, if you think that this greatly complicates the code, I'll close it

PR #4737, opened 2026-06-30
Fix player overlay deck panel visibility toggles

### LINKED ISSUE #4681: Unable to properly hide ETC's band and the cards at the bottom of the deck
### Expected Behavior

I want to hide ETC's band

Image

### Actual Behavior

actually doesn't hide it

Image

_No response_

### Log/Screenshots

_No response_

numbereleven-a:
﻿## Summary

Fixes #4681.

This updates the player overlay layout handling so the visibility toggles for:

- cards above the deck
- cards below the deck
- E.T.C.'s Band

are respected by both the main in-game overlay and the standalone player window.

The checkbox handlers now also rebuild the main overlay layout and refresh the current card lists after a visibility change. This prevents the E.T.C.'s Band panel from staying visible after being disabled, and prevents it from coming back empty when toggled off and on again.

## Validation

Tested in-game with a constructed deck containing E.T.C., Band Manager and a filled E.T.C. band.

Before the fix:

- the E.T.C.'s Band panel was visible in the player overlay;
- disabling the `E.T.C.'s Band` option did not hide the panel;
- after partial layout refresh changes, toggling it off and on could bring the panel back empty.

After the fix:

- disabling `E.T.C.'s Band` hides the panel immediately;
- enabling it again restores the panel with all three band cards;
- repeated off/on toggles keep working correctly;
- the same layout visibility checks are applied to the standalone player window.

Build validation:

- `Hearthstone Deck Tracker.csproj`
- `Debug`
- `x86`

### DIFF
--- a/Hearthstone Deck Tracker/Controls/ElementSorterItem.xaml.cs
+++ b/Hearthstone Deck Tracker/Controls/ElementSorterItem.xaml.cs
@@ -70,17 +70,21 @@ namespace Hearthstone_Deck_Tracker
 				return;
 			_setConfigValue(true);
 			Config.Save();
-			Core.Overlay.Update(false);
 			if(_isPlayerList)
 			{
+				Core.Overlay.UpdatePlayerLayout();
 				Core.Windows.PlayerWindow.Update();
 				Core.Windows.PlayerWindow.UpdatePlayerLayout();
+				Core.UpdatePlayerCards(true);
 			}
 			else
 			{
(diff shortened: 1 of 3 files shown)


beheh:
Hello, thanks for your PR! The skipping in UpdatePlayerLayout is not the right approach, and instead the existing pattern is to set visibilities in `Update` in `OverlayWindow.Update`. I'll use this PR as a basis, but will push a clean version of it.

PR #4738, opened 2026-06-30
Fix Abyssal Curse overlay counter updates

### LINKED ISSUE #4730: Abyssal Curse counter doesn't count
### Expected Behavior

When I use Abyssal Curse the counter shows the maximum current number of damage it deals

### Actual Behavior

It stays always 0

Play Warlock with Abyssal Curse cards. Play some cards with Abyssal Curse effect. The count stays at zero even after some turns.

### Log/Screenshots

Image

numbereleven-a:
## Summary

Fixes the Abyssal Curse overlay counter not updating after Curse-generating Warlock cards are played.

Closes #4730.

## Root cause

The Abyssal Curse counter was only updated from `TAG_SCRIPT_DATA_NUM_1` tag changes on the Curse token entity. In the current game logs, Curse tokens are created and tracked, but the overlay counter does not reliably receive an update through that tag-change path.

HDT already detects Abyssal Curse creation in `PowerHandler` and updates the internal player/opponent Curse count through `HandlePlayerAbyssalCurse` / `HandleOpponentAbyssalCurse`, but that path did not update the overlay counter system.

There was also a side mismatch for the overlay counter: the visible counter represents the player who created the Curse, while the Curse token itself is controlled by the player who received it.

## Fix

- Update the Abyssal Curse overlay counter from the existing Abyssal Curse game event handlers.
- Route the update to the side that created the Curse.
- Keep the existing tag-change path as a fallback, but align it with the same creator-side semantics.

## Validation

Tested against HDT `v1.53.0` / commit `9fa312ba125cefcbb66d6a544613412db6d3713d`.

Before the fix:

- Used the installed HDT `1.53.0.7337`.
- Played a Warlock deck with Abyssal Curse cards.
- After playing a Curse-generating card, the Abyssal Curse overlay counter stayed at `0`.

After the fix:

- Built HDT from `v1.53.0` with this change.
- Launched the fixed debug build.
- Confirmed in the HDT log:
  - `HDT: 1.53.0.0`
  - active deck: `Custom Warlock`
  - Curse token creation: `TSC_955t`
- Confirmed in-game that the Abyssal Curse counter value updates in the overlay after Curse cards are played.

### DIFF
--- a/Hearthstone Deck Tracker/GameEventHandler.cs
+++ b/Hearthstone Deck Tracker/GameEventHandler.cs
@@ -2786,8 +2786,17 @@ namespace Hearthstone_Deck_Tracker
 				card.Info.CostReduction += value;
 		}
 
-		void HandlePlayerAbyssalCurse(int value) => _game.Player.UpdateAbyssalCurse(value);
-		void HandleOpponentAbyssalCurse(int value) => _game.Opponent.UpdateAbyssalCurse(value);
+		void HandlePlayerAbyssalCurse(int value)
+		{
+			_game.Player.UpdateAbyssalCurse(value);
+			_game.CounterManager.HandleAbyssalCurse(false, value);
+		}
+
+		void HandleOpponentAbyssalCurse(int value)
+		{
+			_game.Opponent.UpdateAbyssalCurse(value);
+			_game.CounterManager.HandleAbyssalCurse(true, value);
+		}
 
 		#endregion
 
--- a/Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
+++ b/Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using HearthDb.Enums;
+using Hearthstone_Deck_Tracker.Hearthstone.CounterSystem.Counters;
 using Hearthstone_Deck_Tracker.LogReader.Interfaces;
 
 namespace Hearthstone_Deck_Tracker.Hearthstone.CounterSystem;
@@ -81,6 +82,13 @@ public class CounterManager
 		{
 			opponentCounter.HandleChoicePicked(choice);
 		}
+	}
+
+	public void HandleAbyssalCurse(bool controlledByPlayer, int value)
+	{
+		var counters = controlledByPlayer ? PlayerCounters : OpponentCounters;
+		foreach(var counter in counters.OfType<AbyssalCurseCounter>())
+			counter.UpdateAbyssalCurse(value);
 	}
 
 	public void Reset()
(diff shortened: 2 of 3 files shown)


Hearthstone Deck Tracker/GameEventHandler.cs
```diff
@@ -2827,8 +2786,17 @@ void HandleOpponentHandCostReduction(int value)
...
+		void HandlePlayerAbyssalCurse(int value)
+		{
+			_game.Player.UpdateAbyssalCurse(value);
+			_game.CounterManager.HandleAbyssalCurse(false, value);
+		}
+
+		void HandleOpponentAbyssalCurse(int value)
+		{
+			_game.Opponent.UpdateAbyssalCurse(value);
+			_game.CounterManager.HandleAbyssalCurse(true, value);
```
beheh, line 2798: This callback and it's related invokation aren't actually used at this time, I belive. The modern pattern for counters is also to defer is into the coutner code itself, so the code doesn't have counter calls scattered in various GameEvent handlers.

My proposal: Drop HandlePlayerAbyssalCurse/HandleOpponentAbyssalCurse here, and port code that invokes it (in `PowerHandler.cs`) directly into AbyssalCurseCounter itself. That should be much cleaner.

Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
```diff
@@ -83,6 +84,13 @@ public void HandleChoicePicked(IHsCompletedChoice choice)
 		}
 	}
 
+	public void HandleAbyssalCurse(bool controlledByPlayer, int value)
+	{
+		var counters = controlledByPlayer ? PlayerCounters : OpponentCounters;
+		foreach(var counter in counters.OfType<AbyssalCurseCounter>())
+			counter.UpdateAbyssalCurse(value);
+	}
```
beheh, line 92: If we drop the externally invoked Abyssal Curse invocation an defer directly to the counter, this is not necessary.

beheh (changes requested):
Hey, thanks for your contribution! There's a few good things, from a code quality level the is side switch obivously make sense.

However the counter logic part of this PR needs some improvement: it wires our legacy Abyssal Curse counter implementation (that is currently unused) to the new counter system. That doesn't seem great, because now we have multiple pieces of code for the same logic hooked up, and it's hard to maintain.

I'd suggest the following instead:
1. get rid of the Abyssal Curse handling in `GameEventHandler.cs`, `PowerHandler.cs`, `Player.cs`
2. update the new-style `AbyssalCounter.cs` with fixed logic, which you can borrow from the aforementioned code.

That way we get a nicely maintaineable fix here, rather than wiring up and bypassing conflicting implementations.

Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
numbereleven-a: That makes sense, thanks for the explanation.

Would you be willing to take over and implement this in the new counter system directly? If so, I can close this PR.

mateuscechetto:
Issue was solved in v1.53.6

PR #4739, opened 2026-07-01
Show Aessina counter for player decks with Malorne the Waywatcher

numbereleven-a:
Closes #4710

## Summary

Shows the Aessina friendly-minions-died counter for player decks that include Malorne the Waywatcher.

Malorne can Discover Aessina as one of the Legendary Wild Gods, so the friendly minions died count is relevant before Aessina is actually generated. Without this, the counter only becomes visible after Aessina is created, which is too late to help with the Discover decision.

The counter still displays Aessina as the related card. Malorne is only used as a player-side generator that makes the counter relevant.

Also refreshes visible counters when the active deck changes, so the overlay updates after HDT auto-selects or switches the active deck.

## Validation

- Reproduced on HDT 1.53.2.7343 with a Custom Mage deck containing Malorne the Waywatcher and no Aessina/Starsurge: Aessina counter was not visible in-game.
- Tested the fixed x64 Debug build in-game with the same deck: Aessina counter appeared at the start of the match before Aessina was generated.
- Added regression coverage for showing the player Aessina counter when Malorne is in the active deck.
- Ran `HDTTests` build.
- Ran `FriendlyMinionsDiedThisGameCounterTest.ShouldShowForPlayerDeckWithMalorneTheWaywatcher`.

### DIFF
--- a/HDTTests/Hearthstone/CounterSystem/FriendlyMinionsDiedThisGameCounterTest.cs
+++ b/HDTTests/Hearthstone/CounterSystem/FriendlyMinionsDiedThisGameCounterTest.cs
@@ -1,0 +1,46 @@
+using System.Linq;
+using System.Reflection;
+using Hearthstone_Deck_Tracker;
+using Hearthstone_Deck_Tracker.Hearthstone;
+using Hearthstone_Deck_Tracker.Hearthstone.CounterSystem.Counters;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace HDTTests.Hearthstone.CounterSystem
+{
+	[TestClass]
+	public class FriendlyMinionsDiedThisGameCounterTest
+	{
+		[TestCleanup]
+		public void Cleanup()
+		{
+			SetActiveDeckWithoutEvents(null);
+		}
+
+		[TestMethod]
+		public void ShouldShowForPlayerDeckWithMalorneTheWaywatcher()
+		{
+			SetActiveDeckWithoutEvents(new Deck
+			{
+				Cards =
+				{
+					new Card(HearthDb.CardIds.Collectible.Neutral.MalorneTheWaywatcher)
+				}
+			});
+
+			var counter = new FriendlyMinionsDiedThisGameCounter(true, new GameV2());
+
+			Assert.IsTrue(counter.ShouldShow());
+			CollectionAssert.AreEqual(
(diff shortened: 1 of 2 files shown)


numbereleven-a:
image

Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
```diff
@@ -16,6 +16,7 @@ public class CounterManager
 	public CounterManager(IGame game)
 	{
 		Game = game;
+		DeckList.Instance.ActiveDeckChanged += _ => NotifyCountersChanged();
```
beheh: This change is interesting. Why is it necessary? It makes sense, but existing counters work just fine without it, so I'd like to understand the motivation.

beheh (changes requested):
Thanks for the PR! Code-quality wise this looks, except I'd like to understand why you think the DeckList.Instance.ActiveDeckChanged callback is necessary.

For whether this makes sense gameplay-wise I'd defer to @mateuscechetto.

Hearthstone Deck Tracker/Hearthstone/CounterSystem/CounterManager.cs
numbereleven-a: Good point. The ActiveDeckChanged callback was added because counter visibility depends on DeckList.Instance.ActiveDeck, and I wanted the overlay to refresh if the active deck changes after the counters have already been initialized.

I think you're right, it's not necessary, so I'll delete it?

I don't understand why I need to contact @mateuscechetto.

beheh: Yeah, I think remove that, at least as part of this PR.

What I'm saying is that my feedback so far is preliminary and only for code quality, but ultimately Mateus will take the merge decision and whether this makes sense gameplay-wise.

mateuscechetto:
Hey, I believe gameplay wise this change would make the counter appear all the time whenever malorne is in a deck, which feels unnecessarily noisy. 
With the current behavior, the counter appears after Malorne is played and Aessina is one of the available choices, giving the player the information exactly when it's relevant to deciding whether to pick it.

Closing the PR and the issue because I don't think this change improves the gameplay experience, and I prefer to keep the current, more contextual behavior.

PR #4740, opened 2026-07-01
fix(overlay): replace WPF Buttons with Border+MouseLeftButtonUp in BG…

JasonLandbridge:
… overlays

WPF Button controls do not fire click events reliably under Wine/Linux when
used inside WS_EX_TRANSPARENT overlay windows.

GuidesTabs (top-right tabs): 3 Buttons replaced, fixing unclickable tabs.
BattlegroundsMinionPinning (bottom-left buttons): 6 Buttons replaced.
All 9 use Border+MouseLeftButtonUp matching proven BattlegroundsTierButton pattern.

Added regression tests asserting zero Button elements in overlay XAML.

- [ ] I have read the [Contributing Guidelines](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/CONTRIBUTING.md#contributing)
- [ ] I have already signed the CLA

### DIFF
--- a/.github/workflows/build-portable.yml
+++ b/.github/workflows/build-portable.yml
@@ -19,6 +19,11 @@ on:
         description: "Custom BobsBuddy.zip URL (optional)"
         required: false
         type: string
+      sign_build:
+        description: "Code sign executables with DigiCert secrets"
+        required: false
+        type: boolean
+        default: false
 
 defaults:
   run:
@@ -91,12 +96,14 @@ jobs:
           exit 0
 
       - name: Setup SM_CLIENT_CERT_FILE from base64 secret data
+        if: ${{ inputs.sign_build }}
         shell: bash
         run: |
           echo "${{ secrets.SM_CLIENT_CERT_FILE_B64 }}" | base64 --decode > ${RUNNER_TEMP}/sm_client_cert.p12
           echo "SM_CLIENT_CERT_FILE=${RUNNER_TEMP}/sm_client_cert.p12" >> $GITHUB_ENV
 
       - name: Setup Software Trust Manager
+        if: ${{ inputs.sign_build }}
         uses: digicert/code-signing-software-trust-action@v1.0.1
         with:
           simple-signing-mode: true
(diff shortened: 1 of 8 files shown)


beheh:
Hi there, thanks for your PR! All contributions to HearthSim require a signed CLA. I've sent you an [invite](https://github.com/HearthSim/cla/invitations) to [the repository](https://github.com/HearthSim/cla), please follow the instructions there. If you've already signed the CLA, please let us know.

The board snapshot change looks genuinely useful - could you please pull that into a separate PR?

Regarding the button change: I get the idea with the border and the need for this - many on our team use Linux too. Nonetheless I don't feel super stoked about the idea of dropping the native Buttons, as it feels like it'll make the code base a bit less ergonomic to write. Do you have any more background on this change? Why do Buttons struggle with `WS_EX_TRANSPARENT`, and why does the Border component fix this?

JasonLandbridge:
@beheh 

> Hi there, thanks for your PR! All contributions to HearthSim require a signed CLA. I've sent you an [invite](https://github.com/HearthSim/cla/invitations) to [the repository](https://github.com/HearthSim/cla?rgh-link-date=2026-07-01T09%3A48%3A00Z), please follow the instructions there. If you've already signed the CLA, please let us know.

Hi and yes, I will sign once I'm finished with this PR

> The board snapshot change looks genuinely useful - could you please pull that into a separate PR?

I will, I accidentally made the PR to the upstream repo instead of my own fork to first test it out. I'm trying to have a build and test it locally with Wine to confirm it now works. 

> Regarding the button change: I get the idea with the border and the need for this - many on our team use Linux too. Nonetheless I don't feel super stoked about the idea of dropping the native Buttons, as it feels like it'll make the code base a bit less ergonomic to write. 

Completely understandable, as I made this PR prematurely I was still investigating if this might fix it. From what I could gather, other buttons do work such as the settings button in the overlay but the others do not so this was a "hail mary" fix

> Do you have any more background on this change? Why do buttons struggle with `WS_EX_TRANSPARENT`, and why does the Border component fix this?

The problematic controls were the ones relying on WPF ButtonBase semantics: mouse capture on press, Click firing only after the correct down/up sequence, and command routing through the control. Under Wine/Linux, in this transparent overlay window configuration, that appears to be unreliable: the controls render and hover state can sometimes work, but the actual Click/Command path does not consistently fire.

The reason I tried Border + MouseLeftButtonUp is that HDT already uses that simpler pattern successfully in other interactive Battlegrounds overlay controls. For example, controls like the Battlegrounds tier buttons and hero panel do not use native WPF Button; they listen directly for mouse events and invoke the command manually. Those controls are clickable for me under Wine, while the Button-based ones were not. So the PR applies the same known-working overlay interaction model to the affected tabs/buttons.

That said, I understand the concern about losing native Button ergonomics. A cleaner version might be to introduce a small reusable overlay-specific control/style/helper — something like OverlayButton — that preserves a button-like authoring experience while internally using the more reliable mouse-event path for WS_EX_TRANSPARENT overlays. I’m happy to revise the PR in that direction if that would be preferable.

But, I just tried this out in a Hearthstone battlegrounds Dual match and the tabs work perfectly on first click:
Screenshot_20260701_124246

The red boxed buttons work on first click now, and the green boxed buttons work with double click so I can try that fix there as well. 

Screenshot_20260701_124328

These buttons also work on first click now

JasonLandbridge:
Closing this in favor of: 

https://github.com/HearthSim/Hearthstone-Deck-Tracker/pull/4742
https://github.com/HearthSim/Hearthstone-Deck-Tracker/pull/4741

Issue #3970, opened 2019-10-08
Today's update 1.8.20 causes hearthstone to ignore mouse & keyboard input


foresto:
Before today's HDT update, I was able to play Hearthstone (including the new Hallow's End version 15.4.0.34670) with HDT. Since HDT updated itself to 1.8.20, Hearthstone still launches, but it ignores all mouse and keyboard input,

If I launch Hearthstone standalone instead of via HDT, this does not happen.

I'm playing on Linux via Lutris, in case that matters. I have been doing so for half a year, so I'm surprised to have this problem all of a sudden.

foresto:
Found a Wine registry setting that works around the problem:

    Windows Registry Editor Version 5.00
    [HKEY_CURRENT_USER\Software\Wine\X11 Driver]
    "UseTakeFocus"="N"

I haven't found docs for this settings, but it seems to work.

azeier:
I'm gonna guess this is related to Hearthstone upgrading to a newer unity version.

riQQ:
I'm voting for closing this and continuing in the HDT on Linux mega thread: #2511

azeier:
👍

Issue #2511, opened 2016-05-16
Working on Linux?


wavecycle:
I've been trying to get this working on Linux (Ubuntu 14.04) with no success. I've checked the other threads relating to this and none of them offer a conclusive result.

Has anybody gotten this working..and how please?! Thanks

jleclanche:
What issue are you hitting? Are you using Wine? Which version?

wavecycle:
Thanks for the reply! I'm using Wine and PlayonLinux. 

First stumbling block has been installing Dotnet 4.5. I've tried the quick way, namely to create a new 32 bit wine prefix through PoL (1.7.49 according to[ WineHQ Dotnet page](https://appdb.winehq.org/objectManager.php?sClass=application&iId=2586) ) and then install the dotnet45 component. This retro installs whatever previous versions of dotnet it thinks are required. At the end of this I run the [Dotnet diagnostic tool](https://support.microsoft.com/en-za/kb/2698555) and when i check the versions I get:

- 2.0: SP2 verifies
- 3.0: SP1 verifies
- 3.5: verifies
- 4 Client: verifies
- 4 Full: verifies
- 4.5: Verifies

Then I try run the Hearthstone Deck Tracker.exe in the folder and nothing happens. BTW I have placed the HDT folder inside the prefix's program files folder.

I have also tried the long way round with the dotnet install using wintricks to install from the earliest versions, using the the wine versions as listed in the [ WineHQ Dotnet page](https://appdb.winehq.org/objectManager.php?sClass=application&iId=2586) ). Checking each version along the way to check it is working. At the end I run the file and nothing happens.

I found another link referring to [dotnet installation problems](https://forum.winehq.org/viewtopic.php?t=13088) for another game..and that suggested using winetricks to install gdiplus and vcrun2005. I tried that, tried running the HDT exe in the prefix and nothing happens! Well not nothing..I got a slight flickering that looks like a piece of a cursor but nothing else.

I've also tried running another tracker "Hearthstone Tracker" and it wouldn't work without the dotnet 4.5 install. After the 4.5 install the Hearthstone Tracker install runs successfully but when it tries to start the app afterwards nothing happens.

Any help would be greatly appreciated.

jleclanche:
Try running HDT from the command line with `wine Hearthstone Deck Tracker.exe` and pasting the output.

jleclanche:
[...]

PR #4741, opened 2026-07-01
Fix buttons being unclickable on Linux/Wine

JasonLandbridge:
Related: 
#2511 
#3970
 https://github.com/HearthSim/Hearthstone-Deck-Tracker/issues/2511#issuecomment-4005135086

## Summary

Fixes BG overlay buttons being unclickable under Wine/Linux by replacing WPF `<Button>` elements with a new `OverlayButton` control that avoids `ButtonBase` click routing.

## Problem

The Battlegrounds in-game overlay runs in a `WS_EX_TRANSPARENT` window so clicks pass through to Hearthstone when the cursor isn't over a clickable element. `SetClickthrough(true/false)` dynamically toggles this flag at 60 Hz based on cursor position.

Under Wine/Linux, toggling `WS_EX_TRANSPARENT` off and back on causes WPF's `ButtonBase` click routing to miss the paired `WM_LBUTTONDOWN`/`WM_LBUTTONUP` events it needs. The **hover** still works (the cursor detection runs in managed code via `UpdateHoverable`), but **clicks** never fire.

This is why the settings button (`BattlegroundsSession.BtnOptions`) and tier filter button always worked — they were never WPF Buttons. They use `Border + MouseBinding` which fires on the raw `MouseUp` event without going through `ButtonBase`.

## Solution

`OverlayButton` — a `Border` subclass that exposes `Command`/`CommandParameter` dependency properties and a `Click` event, matching the WPF Button authoring model while using `MouseLeftButtonUp` internally.

```csharp
public class OverlayButton : Border
{
    public ICommand? Command { get; set; }          // DP
    public object? CommandParameter { get; set; }   // DP
    public event RoutedEventHandler? Click;

    protected override void OnMouseLeftButtonUp(MouseButtonEventArgs e)
    {
        // guard: IsEnabled, CanExecute
        Click?.Invoke(this, new RoutedEventArgs());
        command?.Execute(commandParameter);
        e.Handled = true;   // prevents click-through to game
    }
}
```

This follows the exact same pattern as the pre-existing `BattlegroundsTierButton` (`UserControl` + `MouseUp`), extracted into a reusable building block with proper MVVM binding support.

## Changes

- **New:** `OverlayButton.cs` — reusable Border subclass with Command/Click support
- **16 XAML files** — all `<Button>` elements in BG overlay replaced with `<battlegrounds:OverlayButton>`
- **Tests:** `GuidesTabsOverlayCompatibilityTest` — asserts zero WPF Button/ButtonBase elements in all 8 overlay XAML files, catches self-closing tags and ButtonBase subclasses
- **Dead code removed:** implicit `Style TargetType="Button"` resources (no longer apply to anything), unused `Comp_Click` handler, unused usings

## Testing

- [x] Tab switching (Minions/Comps/Heroes) works under Wine/Linux
- [x] Minion pinning expand/toggle/recommend buttons work
- [x] Comp guide back button, pin-all, inspiration button work
- [x] Minion extra-filters keyword buttons work
- [x] Regression test suite passes (8 XAML files verified button-free)
- [x] Hover visual feedback (background changes on mouse-over) preserved
- [x] All existing Command bindings preserved (MVVM layer unchanged)
- [x] All existing Click code-behind handlers preserved (signatures unchanged)

I tried this out in a Hearthstone battlegrounds Dual match and the tabs work perfectly on first click:
Screenshot_20260701_124246

These buttons also work on first click now
Screenshot_20260701_124328

- [ ] I have read the [Contributing Guidelines](https://github.com/HearthSim/Hearthstone-Deck-Tracker/blob/master/CONTRIBUTING.md#contributing)


JasonLandbridge:
@beheh Hi, could you please merge this?

YildizBurhan:
For me the buttons were working in ubuntu (wayland) with v1.55.2 but not anymore in v1.55.3 (ubuntu linux. Running it via steam/proton). Any ideas what might be causing this?

beheh:
Sorry, I've been on vacation, but I should be able to get to this this week.

As for why it would've worked on v1.55.2 but not on .3 I don't have an explanation - this changeset has not landed yet, and while there were some tangentially related changes in .3, none of them should've really enabled/broken this behaviour.

YildizBurhan:
Thanks for the info. Well in the last version i could click on the overlay buttons then click back onto the game and after the second click it would trigger the button and show the new overlay panel. But I might have to debug it and create a new issue.

beheh:
I'm just cleaning this branch up today, and FYI I've dropped that component-level test - I'm not a huge fan of how it was written, and it felt more like a one-time LLM artifact to prove that it did it's job, rather than sensible long-time assertion. It was also very constrained to specific parts of the code base.

Merging this soon.

JasonLandbridge:
@beheh Yeah sure no problem, I'm happy its fixed now! Also appreciate the corrections you made! Thanks!

PR #4758, opened 2026-07-23
Add collection export to file

### LINKED ISSUE #4667: Collection export to file
I'd like a feature for Deck Tracker to export my current synced account card collection to a file (JSON or CSV). Basically just expose same collection data that gets sent to HSReplay.

Current JSON format of api response HSReplay sends when accessing https://hsreplay.net/collection/mine/ is good enough for me. My use case is that I want to make a collection disenchant calculator app that users could easily import their collection into.

This export feature probably can be already done through simple extension if deck tracker exposes necessary API, but it would be much better if that feature was already baked into deck tracker itself.

dokson:
## Summary

Fixes #4667.

Adds an **Export collection to file** button to the HSReplay collection options page. It serializes the currently synced collection to JSON (same format as the HSReplay collection API) through a save-file dialog. The button is shown once the collection has been synced.

When no collection is available (Hearthstone not running / never synced) a message informs the user instead.

## Screen

image

## Validation

- Exported a real collection from the built app and verified the JSON (cards as `dbfId -> [counts]`, `dust`, `favorite_heroes`, `player_records`, …).
- Added unit tests for the serialization contract (top-level keys, account identifiers excluded, premium-type count layout, size).

Build validation:

- `Debug`
- `x86` / `x64`

### DIFF
--- a/HDTTests/Hearthstone/CollectionTest.cs
+++ b/HDTTests/Hearthstone/CollectionTest.cs
@@ -1,0 +1,58 @@
+using System.Linq;
+using System.Text.RegularExpressions;
+using HearthMirror.Objects;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using HDTCollection = Hearthstone_Deck_Tracker.Hearthstone.Collection;
+
+namespace HDTTests.Hearthstone
+{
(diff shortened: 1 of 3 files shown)


beheh:
Hey there, thanks for your contribution! However, you can already share your collection on HSReplay.net and get a JSON from there. We don't think this is necessary as a default feature on HDT.

Issue #4753, opened 2026-07-16
Session MMR/rating overlay not updating in Battlegrounds Duos despite correct data in BgsLastGames.xml


email2morten-spec:
- **HDT version:** 1.53.9.7362 (latest)
- **OS:** Windows 10 Home 26200

**Problem:** In Battlegrounds Duos, the session recap overlay does not show MMR from previous games played earlier in the same session. This does not happen in solo Battlegrounds — session MMR tracking works correctly there.

**What I found while investigating:**
- `BgsLastGames.xml` actually contains correct per-game `Rating`/`RatingAfter` values for `Duos="true"` games, so the underlying rating data is being captured and persisted correctly.
- The current session log contains ~95 repeated errors during Duos play:
  `HearthMirror RPC [client]: Error while invoking GetAccountId: RuntimeBinderException: Cannot perform runtime binding on a null reference`
- These errors are not present (or far less frequent) during solo sessions in the same log history.
- Relevant session recap settings (`ShowSessionRecapStartCurrentMMR`, etc.) are all enabled in `config.xml`.

**Hypothesis:** Since Duos has a teammate present in addition to the player, the `GetAccountId` memory lookup may be resolving ambiguously or failing where it doesn't in solo, which could explain why the live session overlay fails to track/display MMR progression in Duos even though post-game history is saved correctly via a separate path.

**Repro:** Play multiple consecutive Duos Battlegrounds games in one HDT session; session recap does not reflect MMR change from earlier games in that session.

beheh:
Hi, HDT .13 (we just released today) has a potential fix for this. Please let us know either if it's fixed, or if it's still happening to you. Thanks!

beheh:
Considering this fixed. If not, please open a new issue.

PR #4760, opened 2026-07-23
Harden Battlegrounds Duos session recap

dokson:
## Summary

May address #4753.

Two issues in the Battlegrounds session recap, one provable and one defensive:

1. **Solo rating used in Duos (provable).** The end-of-session MMR reset check in `GetSessionGames` compared `BattlegroundsRatingInfo.Rating` (solo) even in Duos. A duos player whose solo rating is low (e.g. unranked/0 this season) could satisfy the "MMR reset" condition and have the whole session recap wiped. It now uses the mode-appropriate rating (`DuosRating` in Duos), matching the current-rating read a few lines above.

2. **Recap wiped on transient account lookup failure (hardening).** The live recap reads the game list via `GetAccountId`, which can transiently return null / throw. That returned an empty list and cleared the recap. `PlayerGames` now returns `null` (rather than an empty list) when the account can't be resolved, and the view model keeps the current recap for that cycle instead of clearing it. The post-game `BgsLastGames.xml` save is unaffected (its rating comes from a separate, null-safe source).

The session computation was extracted into a pure, testable method to enable the tests below.

## Validation

- Could **not** reproduce the reporter's `RuntimeBinderException` locally on 1.53.14 (their report was on 1.53.9.7362), so item 2 is defensive hardening rather than a confirmed fix. Item 1 is a concrete defect independent of that exception.
- Added unit tests (TDD) for the extracted session logic: single-session retention, split on >2h gap, the duos-rating reset check, genuine MMR reset, and stale-session wipe.
- Full test suite green (219 tests).

Build validation:

- `Debug`
- `x86` / `x64`

### DIFF
--- a/HDTTests/Controls/Overlay/Battlegrounds/Session/BattlegroundsSessionTests.cs
+++ b/HDTTests/Controls/Overlay/Battlegrounds/Session/BattlegroundsSessionTests.cs
@@ -1,0 +1,85 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Hearthstone_Deck_Tracker.Controls.Overlay.Battlegrounds.Session;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using static Hearthstone_Deck_Tracker.Utility.Battlegrounds.BattlegroundsLastGames;
+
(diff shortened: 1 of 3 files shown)


beheh:
Thanks for your PR! I think both of these have been since fixed independently.

PR #4761, opened 2026-07-23
Add pre-commit whitespace hook

dokson:
## Summary

Adds a shared git pre-commit hook (`.githooks/pre-commit`) that enforces the existing `.editorconfig` whitespace conventions on **staged changes only**, so existing code is never touched:

- **Auto-fixes** trailing whitespace on staged files and re-stages them, except markdown (line-break semantics) and partially-staged files (to avoid clobbering `git add -p`).
- **Blocks** (does not auto-fix) CRLF and wrong indentation (spaces in code, tabs in xml/xaml) on staged lines. Bypass with `git commit --no-verify`.

Wiring:
- `bootstrap.ps1` sets `core.hooksPath .githooks` (run on setup already).
- `CONTRIBUTING.md` documents the manual opt-in (`git config core.hooksPath .githooks`).
- `.gitattributes` pins the hook to LF so it works when checked out on Unix.

The hook is POSIX `sh` + `awk`/`sed`, works on Unix and Windows Git Bash, and adds nothing to the build.

## Validation

- Synthetic cases: trailing-ws auto-fix + re-stage; space-indentation and tab-in-xaml blocked; markdown trailing-ws preserved; partially-staged files left untouched.
- Real-file check: touching a tracked file cleaned its pre-existing trailing whitespace on the staged content and blocked an introduced space-indented line, then reverted cleanly.

Note: because trailing-ws fixing is whole-file, the first commit that touches a file with pre-existing trailing whitespace will also clean that file (invisible, harmless).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.cs eol=lf
 *.xaml eol=lf
+.githooks/pre-commit eol=lf
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,88 @@
+#!/bin/sh
+# Pre-commit whitespace/style checks aligned with .editorconfig.
+# - Auto-fixes trailing whitespace on staged files and re-stages them,
+#   except markdown (line-break semantics) and partially-staged files.
+# - Blocks (does not auto-fix) CRLF and wrong indentation on staged lines.
+# Only staged content is inspected, so existing code is left untouched.
+# Bypass with: git commit --no-verify
+
+IFS='
+'
+fail=0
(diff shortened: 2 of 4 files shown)


beheh:
Hey, thanks for your PR! However, I'm categorically not a huge fan of pre-commit hooks because in my experience they tend to massively slow down Git workflows such as rebases. I especially don't think a hand-rolled script  is the solution for this either.

I'd rather enforce them at the CI level first and clean up the code-base, or have a command we re-run periodically or can hook up to IDEs.

So closing for now.

PR #4764, opened 2026-07-23
Clean up compiler warnings and dead code

dokson:
## Summary

Broad compiler-warning cleanup across the solution, driven by the analyzer diagnostics and split into three self-contained commits:

1. **Remove dead code and unused usings** (`chore`). Removes compiler-flagged dead code — unused private fields (`CS0169`/`CS0414`), a dead local (`CS0219`), a redundant default parameter on an explicit interface impl (`CS1066`), a duplicate `using` (`CS0105`), and an always-`false` comparison (`CS0472`) — then sweeps **241 unused `using` directives across 182 files** (`IDE0005`). Adds `dotnet_diagnostic.IDE0005.severity = warning` to `.editorconfig` so unused imports are flagged from now on. Also extracts the Battlegrounds `PlayerGames` filter into a pure, testable method.

2. **Log swallowed exceptions instead of ignoring them** (`refactor`). Six `catch` blocks that silently swallowed (`// ignored`, empty, or `#if DEBUG`-only) now log at debug level.

3. **De-duplicate Arena hero-option view models** (`refactor`). `ArenaPickSingleDualClassHeroOptionViewModel` and `ArenaPickSingleHeroPowerOptionViewModel` were near-verbatim copies of `ArenaPickSingleHeroOptionViewModel` (identical properties, even a copy-pasted comment), differing only by a per-position margin. They now inherit the shared base — which takes an optional `plaqueMargin` — and `ArenaPlaqueViewModel`'s two constructors collapse into one with an optional margin. This clears the `CS0108` member-hiding warnings. Behavior is preserved: WPF binds by name, so inherited members resolve identically.

Files already touched by the other open PRs (#4758, #4759, #4760, #4761) were deliberately excluded from the sweep to avoid conflicts.

## Validation

- Full test suite green (**221 tests**), including new tests added here: the `PlayerGames` duos/solo/player filter, and an Arena margin regression test locking each view model's per-position `Margin` and `PlaqueViewModel.Margin` across both constructors.
- 0 build errors; 0 remaining unused imports (`IDE0005`) outside the excluded files.

Build validation:

- `Debug` (x64)


dokson:
@mateuscechetto can you have a look at this refactor?

Issue #3445, opened 2017-10-18
Deck list is offscreen on 2nd display


SOLiNARY:
### Expected Behavior
To be on one display where the game is.

### Actual Behavior
Deck list ingame is offscreen on 2nd display.

### Steps to reproduce behavior
Update Hearthstone, change language to English (US), Update HDT, launch both.

### Log/Screenshots
![issue_hdt](https://user-images.githubusercontent.com/3579801/31706721-0051b5e2-b3fb-11e7-8835-f652c3445b59.jpg)

MrMadden:
When will this get fixed? I like to play arena and watch netflix but now I can't use the deck tracker because its in front of whatever I have on the second screen

azeier:
Does the resolution set in Hearthstone match the screen resolution?
Does the issue remain if Hearthstone is in Windowed mode?

MrMadden:
Yes and yes

SOLiNARY:
in windowed mode its ok, but fullscreen is wrecked.

SOLiNARY:
@azeier if u need more info just ask, this issue is still actual.

MrMadden:
Yea this is still an issue. I don’t even use HDT anymore due to this bug.

duiker2:
Any update on this? I'm also experiencing this bug.

duiker2:
I just found a temporary workaround. It seems to be happening when the screen scale for one monitor is different from the other (in Windows 10 Settings -> System -> Display -> Screen Scale). Changing them to the same screen scale % at least while playing hearthstone fixes the issue.

dominics:
@duiker2 is completely correct from my testing. I have three monitors, scaled 150% 150% and 100% (2 UHD, 1 1080p), and setting them all to any consistent screen scaling seems to work. Furthermore, turning on this compatibility option ("override high DPI scaling behaviour") seems to force the issue and allow me to use mixed scaling (requires a restart of HDT to take effect)

![image](https://user-images.githubusercontent.com/97427/33515943-d792b622-d7ce-11e7-8dd0-c38371aa0f8d.png)

MrMadden:
@dominics thank you man! That bug was so frustrating to me

strican:
[...]

PR #4773, opened 2026-08-22
[Investigation, likely insufficient] Opponent deck/cards off-screen on mixed-DPI monitors

infame:
## Status
Opening as **draft**: this addresses one concrete, verified root cause (see below), but I have not been able to runtime-test it (no Windows/multi-monitor mixed-DPI rig available to me), and there's a real chance it only fixes part of the underlying problem - see "Known limitation" below. Marking as draft rather than claiming this is a complete fix.

## Summary
- `GetHearthstoneRect()` converts Hearthstone's client rect from screen pixels to WPF units using the global `Helper.DpiScalingX`/`DpiScalingY` fields.
- Those fields are set exactly once each, in `MainWindow`'s and `OverlayWindow`'s `Loaded` handlers, from whichever window's monitor happened to be current *at load time* - not from the monitor Hearthstone is actually on, and never refreshed afterwards (no `WM_DPICHANGED`/`DpiChanged` handling anywhere in the codebase). This part I verified directly by reading the code - it's unambiguously stale/wrong-window state.
- On a multi-monitor setup with mixed scaling (e.g. HDT's windows first load on a 100% monitor, Hearthstone runs fullscreen on a 150% monitor), this produces a wrong client rect. `OverlayWindow.UpdatePosition()` applies that rect directly to the whole overlay's `Top/Left/Width/Height`, so the overlay - including the opponent's card/deck panel - ends up shifted, sometimes off the visible game area.
- This matches the long-standing community report in #3445 ("Deck list is offscreen on 2nd display"), where the only known workaround was setting all monitors to the same scaling, or forcing "system DPI" override in the exe's compatibility settings.

## Fix
- Use `GetDpiForWindow` (Windows 10 1607+) to read the DPI of the monitor Hearthstone's own window is currently on, at the point we compute its rect, instead of relying on stale global state captured from an unrelated window.
- Falls back to the previous `Helper.DpiScalingX/Y` behavior if the API isn't available (e.g. older Windows versions HDT still supports), so there's no regression on unsupported OS versions.
- Scoped narrowly to `User32.GetHearthstoneRect`; no change to the app's overall DPI-awareness mode, to keep the change low-risk.

## Known limitation
In the #3445 thread, `@strican` found evidence that in Hearthstone's **fullscreen** mode specifically, `GetClientRect`/`ClientToScreen` may return coordinates as if DPI virtualization weren't happening at all (possibly a Unity/DirectX fullscreen quirk), independent of which monitor's DPI you divide by. If that's still the case, this PR fixes the "wrong/stale monitor" half of the bug but may not fully fix the fullscreen case by itself. I don't have a way to verify which of these is actually happening on current Windows/Hearthstone versions - flagging this so reviewers/testers know what to check for.

## Request for verification
I'd appreciate it if someone who can reproduce #3445, or a maintainer, could test this on a real mixed-DPI multi-monitor setup (both windowed and fullscreen Hearthstone) before this is considered for merge. Happy to adjust the approach (e.g. `MonitorFromWindow`/`GetDpiForMonitor`) if needed.

I'm also aware `CONTRIBUTING.md` asks for prior approval on non-trivial changes and a signed CLA before merge - opening this as a draft/concrete starting point for discussion rather than a merge-ready PR; happy to go through those steps as needed.

## Test plan
- [ ] Reproduce #3445 on a real mixed-DPI multi-monitor Windows setup with Hearthstone fullscreen on the differently-scaled monitor.
- [ ] Confirm the opponent (and player) overlay panels now render within the visible game area with this change.
- [ ] Repeat in windowed mode.
- [ ] Confirm no regression on a single-monitor / uniform-DPI setup.

### DIFF
--- a/Hearthstone Deck Tracker/Utility/User32.cs
+++ b/Hearthstone Deck Tracker/Utility/User32.cs
@@ -97,6 +97,10 @@ namespace Hearthstone_Deck_Tracker
 		[DllImport("user32.dll")]
 		public static extern bool PrintWindow(IntPtr hwnd, IntPtr hdcBlt, uint nFlags);
 
+		// Only available on Windows 10 1607+; callers must handle EntryPointNotFoundException.
+		[DllImport("user32.dll")]
+		private static extern uint GetDpiForWindow(IntPtr hWnd);
+
 		[DllImport("user32.dll")]
 		public static extern uint GetWindowThreadProcessId(IntPtr hWnd, out uint lpdwProcessId);
 
@@ -272,13 +276,41 @@ namespace Hearthstone_Deck_Tracker
 
 			if(dpiScaling)
 			{
-				ptUL.X = (int)(ptUL.X / Helper.DpiScalingX);
-				ptUL.Y = (int)(ptUL.Y / Helper.DpiScalingY);
-				ptLR.X = (int)(ptLR.X / Helper.DpiScalingX);
-				ptLR.Y = (int)(ptLR.Y / Helper.DpiScalingY);
+				var (scaleX, scaleY) = GetDpiScalingForWindow(hsHandle);
+				ptUL.X = (int)(ptUL.X / scaleX);
+				ptUL.Y = (int)(ptUL.Y / scaleY);
+				ptLR.X = (int)(ptLR.X / scaleX);
+				ptLR.Y = (int)(ptLR.Y / scaleY);
 			}
 
 			return new Rectangle(ptUL.X, ptUL.Y, ptLR.X - ptUL.X, ptLR.Y - ptUL.Y);
+		}
+
+		// Helper.DpiScalingX/Y reflect the DPI of whichever HDT window last raised its
+		// Loaded event, which is not necessarily the monitor Hearthstone is on. On a
+		// multi-monitor setup with mixed scaling this produced the wrong client rect for
+		// the game window (HDT#3445 - opponent's cards/deck rendered off-screen).
+		// GetDpiForWindow (Windows 10 1607+) returns the DPI of the monitor the given
+		// window actually is on, so we prefer that and only fall back to the older,
+		// window-agnostic value on OS versions where the API doesn't exist.
+		private static (double X, double Y) GetDpiScalingForWindow(IntPtr hWnd)
+		{
+			try
+			{
+				var dpi = GetDpiForWindow(hWnd);
+				if(dpi > 0)
+				{
+					var scale = dpi / 96.0;
+					return (scale, scale);
+				}
+			}
+			catch(EntryPointNotFoundException)
+			{
+			}
+			catch(DllNotFoundException)
+			{
+			}
+			return (Helper.DpiScalingX, Helper.DpiScalingY);
 		}
 
 		public static void BringHsToForeground()


infame:
Plain-English version of the above for anyone who doesn't want to read the code diff:

- HDT draws its overlay (opponent's cards, deck counters, etc.) by figuring out exactly where Hearthstone's window is on your screen, then drawing on top of it.
- To do that correctly on Windows, it needs to know the "DPI scale" of the monitor Hearthstone is on (that's the 100% / 125% / 150% setting in Windows Display settings).
- **The bug:** HDT only ever checked that scale *once*, when its own tracker window first opened - and it used whatever monitor *that* window happened to be on at the time. It never checked again after that.
- So if you have multiple monitors set to different scaling percentages, and Hearthstone ends up running fullscreen on a different monitor than the one HDT grabbed its scale from, HDT does its math with the wrong number. That's what pushes the overlay (including the opponent's deck panel) into the wrong spot, sometimes far enough that it's off-screen entirely.
- **The fix:** instead of remembering one scale value forever, HDT now asks Windows "what's the DPI scale of the monitor Hearthstone's window is *actually on right now*" every time it repositions the overlay. So it always uses the correct number for wherever the game currently is, instead of a stale guess from startup.
- **Still unverified:** I don't have a Windows PC with mixed-DPI multiple monitors to actually watch this fix work, so this is a "should fix it based on reading the code very carefully" situation, not a "confirmed fixed" one - see the Known Limitation note above. If you're one of the people hitting this, testing the branch and reporting back here would genuinely help get this merged.

infame:
## Correction - the core premise of this PR is likely wrong, please hold off testing

I need to walk this back. After more digging I found the actual MS docs for `GetDpiForWindow` (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdpiforwindow), and its return value depends on the *process's* DPI_AWARENESS:

| DPI_AWARENESS | Return value |
|---|---|
| UNAWARE | fixed 96 |
| SYSTEM_AWARE | **the system DPI** (one value for the whole process) |
| PER_MONITOR_AWARE | the DPI of the monitor the window is actually on |

HDT doesn't declare Per-Monitor DPI awareness anywhere (no app manifest, no `<System.Windows.Media.MultiMonitorDpiAware>`-style App.config switch), so as a WPF/.NET Framework app it defaults to System DPI Aware. That means `GetDpiForWindow(hsHandle)` in this PR will return the same single system DPI value no matter which monitor Hearthstone's window is actually on - it does **not** give us live per-monitor DPI the way I assumed it would.

In other words: this change likely behaves the same as the existing code for the actual cross-monitor mismatch scenario from #3445, since both end up reflecting "system DPI" rather than the real DPI of Hearthstone's monitor. The "stale value from whichever window loaded first" observation in the PR description is still accurate, but fixing *that* alone probably isn't enough to fix the reported symptom, since the real per-monitor value was never being read to begin with in either version.

A fix that actually reads Hearthstone's real per-monitor DPI would need the app (or at least the relevant window/thread) to be Per-Monitor (V2) DPI aware, which is a much bigger, riskier change - it affects rendering scale, hit-testing, and DPI-change handling across every window, and isn't something I can respons­ibly propose without a way to test it on real hardware.

Marking this as not-ready and will follow up with the maintainers/OP on what to do with it (likely close, or repurpose as a writeup of the investigation for whoever picks up an actual Per-Monitor V2 migration). Sorry for the noise, and thanks to anyone who was about to spend time testing this - please don't yet.

infame:
Closing this. Following up on my correction above: I checked `GetDpiForMonitor`'s docs too, and it has the exact same limitation as `GetDpiForWindow` - for a `PROCESS_SYSTEM_DPI_AWARE` app (which is what HDT is, by default, with no manifest declaring otherwise), *any* Win32 "give me this monitor's DPI" call returns the single system-wide DPI, not the real DPI of the specific monitor you asked about. So there's no small/localized API swap that fixes this - reading Hearthstone's real per-monitor DPI correctly would require the app (or at least the calling thread, e.g. via `SetThreadDpiAwarenessContext`) to actually become Per-Monitor DPI aware, which is a substantially bigger change (affects rendering scale and DPI-change handling across every window) that I can't respons­ibly author and validate without real Windows hardware to test on.

I'd rather close this than leave an unverified/likely-non-functional fix open. Sorry again for the noise on #3445 - the diagnosis of "a stale, wrong DPI value gets used" still seems right, but the actual fix needs someone with a reproducing setup (or a maintainer scoping a proper Per-Monitor V2 migration) to take it from here. Thanks for reading through this.

Issue #3427, opened 2017-09-22
Restore the window position if it is completely off screen


riQQ:
See https://blogs.msdn.microsoft.com/davidrickard/2010/03/08/saving-window-size-and-location-in-wpf-and-winforms/

dwarfnose:
Or you click in the taskbar with the right mouse button and select "move window" (or how it's called in your language). Then use the arrow keys on your keyboard to move the off screen window back into your current window.
(I know this trick cause some java programs cause same problem while using two monitors)

Zawodowiec1532:
in Windows 10 it's impossible ;x Atleast I can't reproduce this. Even with arrows.

riQQ:
@Zawodowiec1532 you can't get the window back or you can't get HDT completely off-screen?

Zawodowiec1532:
@riQQ I cant get window completely off-screen. While using arrows trick it's moving with mouse and it's stopping on edge of screen

riQQ:
Maybe it happens in a multi screen setup, when you move HDT to a secondary screen, close HDT to save the position, then disable this screen and then starting HDT again.

riQQ:
A user reported this in discord happening to the stats window. Maximizing / moving solved it.

![unknown](https://user-images.githubusercontent.com/3404787/35122505-5011672a-fc9f-11e7-8107-378bf87803ac.png)

infame:
Opened #4774 - turns out `PlayerWindow`/`OpponentWindow`/`MainWindow`/`TimerWindow` already have exactly this recovery check (snap back to 100,100 if no screen contains the window), `StatsWindow` was just missing it. Small, mechanical fix - reuses the existing pattern rather than introducing anything new.

beheh:
Closed by #4774

PR #4774, opened 2026-08-22
Recover StatsWindow position when it ends up off-screen

infame:
- [x] I have already signed the CLA (see https://github.com/HearthSim/cla/pull/76)

## Summary
- `PlayerWindow`, `OpponentWindow`, `MainWindow` and `TimerWindow` all restore their saved `Left`/`Top` from config, then check the window's title bar corners against `Screen.AllScreens` and reset to `(100, 100)` if none of them land on a currently connected screen.
- `StatsWindow` restores its saved position the same way, but never got this check. If the monitor it was last on gets disconnected/disabled/reconfigured (or the saved coordinates otherwise stop matching any current screen), it can end up completely off-screen with no way back short of editing/deleting the config file.
- This is #3427, and matches a case a maintainer (@riQQ) mentioned in that thread happening to the stats window specifically (reported via Discord, with a screenshot).

## Fix
Adds the same title-bar-corner-vs-`Screen.AllScreens` check already used (and presumably already battle-tested) in the other four windows - copied verbatim, no new logic. Scoped to `StatsWindow.xaml.cs` only.

## Why I'm fairly confident in this one
Unlike my previous PR on the DPI/mixed-monitor overlay issue (which I ended up closing after finding my fix for it didn't actually work), this isn't a new mechanism - it's applying an existing, already-shipped pattern from 4 sibling windows to a 5th window that's simply missing it. I don't have a Windows box to click through the exact repro (disconnect a monitor HDT was last positioned on, relaunch), but the logic itself isn't new or unverified, it's copy-pasted from code that's already running in production for the other windows.

## Test plan
- [ ] Move the stats window (open as a separate window, not the flyout) onto a secondary monitor, close HDT, disable/disconnect that monitor, relaunch HDT, open the stats window again - it should now appear at (100, 100) instead of staying off-screen.
- [ ] Confirm normal open/close/position-save behavior is unaffected on a single-monitor setup.

### DIFF
--- a/Hearthstone Deck Tracker/Windows/StatsWindow.xaml.cs
+++ b/Hearthstone Deck Tracker/Windows/StatsWindow.xaml.cs
@@ -1,7 +1,10 @@
 ﻿#region
 
 using System.ComponentModel;
+using System.Linq;
 using System.Windows;
+using System.Windows.Forms;
+using Point = System.Drawing.Point;
 
 #endregion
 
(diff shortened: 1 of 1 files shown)


beheh:
Hi there, thanks for your PR! All contributions to HearthSim require a signed CLA. I've sent you an [invite](https://github.com/HearthSim/cla/invitations) to [the repository](https://github.com/HearthSim/cla), please follow the instructions there. If you've already signed the CLA, please let us know.

Once that's done and the CLA has been merged, we'll try to get this merged here as soon as possible.

infame:
@beheh Signed the CLA: https://github.com/HearthSim/cla/pull/76

beheh (approved):

